PR #8, opened 2018-10-30
[MSHADE-303] Suppress module-info.class warning if the file if filtered

danilcha:
Maven always shows “Discovered module-info.class. Shading will break its strong encapsulation.” warning, even then the user has explicitly filtered those files out. The same with “META-INF/INDEX.LIST” file. 

This pull request fixes this by applying file filters BEFORE these built-in warning checkers.


jira-importer:
Resolve #305

PR #11, opened 2018-12-15
[MSHADE-307] adding useDefaultConfiguration flag in ShadeMojo

rmannibucau:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MSHADE) filed 
       pulling in other changes.
 - [X] Each commit in the pull request should have a meaningful subject line and body.
 - [X] Format the pull request title like `[MSHADE-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
 - [X] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

- [X] I'm already an ASF committer

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -40,7 +40,9 @@ import org.apache.maven.plugins.shade.filter.SimpleFilter;
 import org.apache.maven.plugins.shade.pom.PomWriter;
 import org.apache.maven.plugins.shade.relocation.Relocator;
 import org.apache.maven.plugins.shade.relocation.SimpleRelocator;
+import org.apache.maven.plugins.shade.resource.ManifestResourceTransformer;
 import org.apache.maven.plugins.shade.resource.ResourceTransformer;
+import org.apache.maven.plugins.shade.resource.ServicesResourceTransformer;
 import org.apache.maven.project.DefaultProjectBuildingRequest;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.MavenProjectHelper;
@@ -76,8 +78,10 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.Set;
 
 /**
@@ -490,7 +494,7 @@ public class ShadeMojo
                         replaceFile( finalFile, testJar );
                         testJar = finalFile;
                     }
-                
+
                     renamed = true;
                 }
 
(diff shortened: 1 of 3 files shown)


eolivelli (changes requested):
Very useful!

I am sure about the name. A 'default' configuration which is not active by default sounds weird.

What about about 'safe', 'common', 'standard'....

rmannibucau:
Common sounds a good alternative but too generic, I thought about "auto" but was not that happy too. Maybe "autoConfigure" + adding a spi on transformers and filters would make more sense?

eolivelli:
autoConfigure + SPI will be great, but it a more complex change.

The other way would be to have such "useDefaultConfiguration", turned *on* by default, but with a bump in the major version for the plugin.

@rfscholte  what do you think ?

rfscholte:
In general I'd like prefer to keep the number of parameters to a minimum. So if we want to do this, why not set these as defaults. When there are explicit filters/transformers, these will be used. Any empty list would mean: no filters or transformers. And this implies there should be an explicit filter for signatures.

rfscholte:
The new version will be 3.n+1.0. The first 3 means that is requires at least Maven 3. We discovered that this pattern makes is very clear when certain plugins can be used.
We should add this change to the index.html file too, just to be clear. (in the end the result should be the same, because if filters and transformers are specified, nothing will change)

rmannibucau:
Hmm, i thought of that - active by default and ignored if set by the user - but it is rare it is sufficient and does need a log4j, openwebbeans or other transformer so i preferred the composition. Should we have it with 3 values: default = use, skip and merge?

eolivelli:
Okay for versioning and index.html
I can take of take (with a little guidance).

For use/skip/merge....things are getting complicated and complicated.

We should default to 'merge', but will 'use' be useful? As 'merge' is a super set of 'use'

rmannibucau:
Hmm it must be combinable or it is useless and config stays complex.

Merge (default), override, skip?

rfscholte:
Sorry, but it will be default or override: easy to understand and maximum control. If you want full merge/override-control, you need to do that with the magic attributes as described at https://blog.sonatype.com/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/

rmannibucau:
@rfscholte this is my intention but here it does not work cause the config is implicit, this is why i proposed 3 states. Now if you say the impl is ok to turn on by default it can be enough for the next release (and we enhance later if needed). If so, do you have a name preference?

rfscholte:
> If so, do you have a name preference?

for what?

rmannibucau:
Guys, I added the SPI and removed the toggle, can you have a look and let me know if it looks better this way? I like this solution cause we can plug the SPI in log4j, openwebbeans, and other extensions

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -762,55 +766,72 @@ private File resolveArtifactSources( Artifact artifact )
 
     private List<ResourceTransformer> getResourceTransformers()
     {
+        final List<ResourceTransformer> resourceTransformers = new LinkedList<>( getDefaultResourceTransformers() );
         if ( transformers == null )
         {
-            return Collections.emptyList();
+            return resourceTransformers;
```
rfscholte: You're still merging. Here you should replace `Collections.emptyList()` with `getDefaultResourceTransformers()`

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -762,55 +766,72 @@ private File resolveArtifactSources( Artifact artifact )
...
         }
+        resourceTransformers.addAll( Arrays.asList( transformers ) );
+        return resourceTransformers;
+    }
 
-        return Arrays.asList( transformers );
+    private List<ResourceTransformer> getDefaultResourceTransformers()
+    {
```
rfscholte: Since we're not merging, so you can return a list with these 2 transformers.
Also document this on the `transfomers`-parameter.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -823,6 +844,50 @@ private File resolveArtifactSources( Artifact artifact )
             }
         }
 
+        // first check for backward compatibility this is not already configured explicitly
+        boolean addExclusion = true;
```
rfscholte: We need a similar trick here:

    public Collection<Filter> getFilters()
    {
      if ( this.filters == null )
      {
        return defaultFilters();
      }
      else
      {
        return this.filters;
      }
    }

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -823,6 +844,50 @@ private File resolveArtifactSources( Artifact artifact )
...
+            {
+                artifacts = getArtifactIds();
+            }
+            simpleFilters.add( new SimpleFilter(
+                    getMatchingJars( artifacts , new ArtifactId( "*:*" ) ),
+                    Collections.<String>emptySet(),
+                    new HashSet<String>( Arrays.asList( "META-INF/*.SF", "META-INF/*.DSA", "META-INF/*.RSA" ) ) ) );
+        }
+
+        for ( final Filter filter : ServiceLoader.load( Filter.class ) )
```
rfscholte: Using ServiceLoader requires a separate JIRA issue, and I wonder if this is going to work. See https://issues.apache.org/jira/browse/MNG-6275

rfscholte (changes requested):

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rmannibucau: this is the case, resourceTransformers is initialized with getDefaultResourceTransformers. If the else skips the default transformers the feature is pointless since in 80% of the case you need to add a custom transformer or customize a config (like mainClass for the manifest)

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rmannibucau: guess it is addressed by previous comment, we want to merge ;)

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rmannibucau: likely the same

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rfscholte: Let's agree to disagree

rmannibucau: I'm fine to disagree but I'd like to solve the issue which aims to reduce the entry cost of the plugin and save hours of debug time due to a bad packaging which is hard to identify - and even more with jpms since module-info.java drops SPI registration :s.

any alternative proposal to have a trivial to enable good auto configuration compatible with user customizations?

rfscholte: Can't think of a clean, clear and simple solution yet. So let's split it into 2 issues: 1. have good default transformers; 2. combine default+custom transfomers

rmannibucau: Ok, let's split. Just to ensure I get it and not do work for nothing, it means 3 PR - no order:

1. defaults (transformers and filters)
2. SPI
3. combine
?

rmannibucau: 1 done at https://github.com/apache/maven-shade-plugin/pull/12

jira-importer:
Resolve #308

jira-importer:
Resolve #308

rmannibucau:

PR #12, opened 2018-12-19
[MSHADE-307] defaults for transformers and filters

rmannibucau:

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -40,7 +40,9 @@ import org.apache.maven.plugins.shade.filter.SimpleFilter;
 import org.apache.maven.plugins.shade.pom.PomWriter;
 import org.apache.maven.plugins.shade.relocation.Relocator;
 import org.apache.maven.plugins.shade.relocation.SimpleRelocator;
+import org.apache.maven.plugins.shade.resource.ManifestResourceTransformer;
 import org.apache.maven.plugins.shade.resource.ResourceTransformer;
+import org.apache.maven.plugins.shade.resource.ServicesResourceTransformer;
 import org.apache.maven.project.DefaultProjectBuildingRequest;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.MavenProjectHelper;
@@ -76,6 +78,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


eolivelli:
What about adding some description about new default in this page ?
https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html

eolivelli (approved):
good.
CI logs:
https://builds.apache.org/job/maven-box/job/maven-shade-plugin/job/MSHADE-307/

eolivelli:
@rmannibucau tests are failing on Jenkins
https://builds.apache.org/job/maven-box/job/maven-shade-plugin/job/MSHADE-307/1/

rmannibucau:
Another path separator issue :(, fixed hopefully

eolivelli:
good
https://builds.apache.org/job/maven-box/job/maven-shade-plugin/job/MSHADE-307/2/

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -764,12 +767,44 @@ private File resolveArtifactSources( Artifact artifact )
...
+            return getDefaultResourceTransformers();
         }
 
         return Arrays.asList( transformers );
     }
 
+    private List<ResourceTransformer> getDefaultResourceTransformers()
+    {
+        final List<ResourceTransformer> transformers = new LinkedList<>();
+        if ( missTransformer( ServicesResourceTransformer.class ) )
```
rfscholte: `getDefaultResourceTransformers()` is only called when transformers is `null`, so no need for `missTransformer()`-call

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -822,6 +831,16 @@ private File resolveArtifactSources( Artifact artifact )
...
+            getLog().debug( "Adding META-INF/*.SF, META-INF/*.DSA and META-INF/*.RSA exclusions" );
+
+            Map<Artifact, ArtifactId> artifacts = getArtifactIds();
+            simpleFilters.add( new SimpleFilter(
+                    getMatchingJars( artifacts , new ArtifactId( "*:*" ) ),
+                    Collections.<String>emptySet(),
+                    new HashSet<>( Arrays.asList( "META-INF/*.SF", "META-INF/*.DSA", "META-INF/*.RSA" ) ) ) );
+        }
 
         filters.addAll( simpleFilters );
```
rfscholte: I would expect a NPE here with the default filters.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -822,6 +831,16 @@ private File resolveArtifactSources( Artifact artifact )
                 simpleFilters.add( new SimpleFilter( jars, filter.getIncludes(), filter.getExcludes() ) );
             }
         }
+        else if ( this.filters == null )
+        {
+            getLog().debug( "Adding META-INF/*.SF, META-INF/*.DSA and META-INF/*.RSA exclusions" );
```
rfscholte: Please create a method getDefaultFilters()

src/test/java/org/apache/maven/plugins/shade/mojo/ShadeMojoTest.java
```diff
@@ -60,6 +64,50 @@
 public class ShadeMojoTest
     extends PlexusTestCase
 {
+    public void testDefaultConfiguration() throws Exception
```
rfscholte: Very creative with reflection, but I wonder if we are testing the right things. If it is too hard to verify it with a unittest, why not switch to an integration test?

rfscholte (changes requested):

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rmannibucau: not sure i'm following, this is not this.filters but the list created in the method

src/test/java/org/apache/maven/plugins/shade/mojo/ShadeMojoTest.java
rmannibucau: Thought about it but I found it saner as a compromise cause this is really more a whitebox test to check we mutate the user config as expected than an IT which would be more complicated with fewer guarantee on what we do at that level IMHO (here we really want to check the default config is setup and not smoke test it, at least was the intention)

rmannibucau:
Updated the PR, I think the test is more relevant for this particular case than an IT since we don't care how it runs but we care about what it runs, let me know if it is still an issue.

elharo (commented):
Is this still under consideration/development or should we close it?

rmannibucau:
Ca, depend on the SPI PR, could be replaced by a SPI module IMHO.

jira-importer:
Resolve #308

rmannibucau:

PR #15, opened 2019-02-01
[MSHADE-223] - Endless processing with promoteTransitiveDependencies

dmitri-gb:
Fix the endless processing that was caused by not checking the classifier of the dependencies when adding excludes.

I opted for just comparing the results of the `getId` method, as it already contains all the necessary info (groupId + artifactId + type + classifier). The previous check `dep.getType() == null` seems unnecessary, as the type is generally never `null`.

I also added an integration test that reproduces the issue. It includes an `invoker.properties` file that sets `invoker.timeoutInSeconds=60`, because without the fix the issue manifests as a non-terminating build.

PS. This fix likely also addresses the complaints that have been filed under MSHADE-148 after its closure.


xhumanoid:
@khmarbaise @rmannibucau 

Hi, any updates when it will merged? 

without this fix we have infinity loops with 2 artifact but different classifiers

```
        <dependency>
            <groupId>groupId</groupId>
            <artifactId>artifactId</artifactId>
            <version>0.0.1</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>groupId</groupId>
            <artifactId>artifactId</artifactId>
            <version>0.0.1</version>
            <classifier>tests</classifier>
            <scope>test</scope>
        </dependency>
```

rmannibucau:
LGTM, i dont have a computer handy but no blocker to merge it in the week for me

rfscholte:
Fixed with https://github.com/apache/maven-shade-plugin/commit/038d807b38115ff093bf16f3b0a2db439202bf67 and https://github.com/apache/maven-shade-plugin/commit/be330db3b6ecb4d1ac111c991c75399310378e2f
Thanks for the PR!

jira-importer:
Resolve #584

PR #16, opened 2019-03-05
[MSHADE-148] Don't attach exclusions that are already present

philipl:
Although I don't fully understand the scenarios where we re-evaluate
the exclusions of a dependency, I do know that if we do, we will add
them again, leading to a false indication that the dependency list has
changed, leading to a regeneration and re-evaluation of the list,
which then continues indefinitely, as the the same exclusions are
added again and again.

To address the leaf problem, let's ensure that we never add an
exclusion that is already present. That way, even if reevaluation
occurs, it will not result in changes.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -1178,9 +1178,17 @@ public class ShadeMojo
                                 Exclusion exclusion = new Exclusion();
                                 exclusion.setArtifactId( n3.getArtifact().getArtifactId() );
                                 exclusion.setGroupId( n3.getArtifact().getGroupId() );
-                                dep.addExclusion( exclusion );
-                                modified = true;
-                                break;
+                                // only add an exclusion if it's not already present.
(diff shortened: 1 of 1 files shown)


philipl:
addExclusion is generated code in core maven. That’s also why Exclusions don’t have an equals() implementation. I’d add a test case if I could come up with a repro case that triggered on its own. I added some discussion in the JIRA.

philipl:
Note, that the fix in #15 looks like the fix for why exclusions get reevaluated. Even if that is merged, this fix should be taken too for correctness.

maobaolong (approved):
I'm +1 with this PR, hope to merge this before next release. I've patched and tested this PR, it can worked without stuck.

elharo (changes requested):
Still needs a test

gnodet:
I also have this issue when building Apache Camel with a parallel build, similar to the above example.
Given this looks like a threading issue, it's not always reproducible, so it's quite hard to think about a test which could reliably exercise the problem.

jira-importer:
Resolve #505

jira-importer:
Resolve #505

PR #17, opened 2019-03-28
[MSHADE-313] - Keep Services

mkarg:
Closes [MSHADE-313](https://issues.apache.org/jira/projects/MSHADE/issues/MSHADE-313) by providing new option `<keepServices>true</keepServices>` (default is `false`) to keep service classes from getting stripped.

### DIFF
--- a/src/it/MSHADE-313/verify.bsh
+++ b/src/it/MSHADE-313/verify.bsh
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import java.io.*;
+import java.util.jar.*;
+
+String[] wanted =
+{
+    "Main.class",
+    "META-INF/services/SomeServiceInterface",
+    "SomeServiceInterface.class",
+    "SomeServiceClass.class",
+    "SomeReferencedClass.class",
+    "META-INF/services/DependencyServiceInterface",
+    "DependencyServiceInterface.class",
+    "DependencyServiceClass.class",
+    "DependencyReferencedClass.class"
+};
+
+String[] unwanted =
+{
+    "UnusedServiceInterface.class",
+    "UnusedServiceClass.class",
+    "SomeUnreferencedClass.class",
+    "DependencyUnreferencedClass.class"
+};
+
+JarFile jarFile = new JarFile( new File( basedir, "test/target/test-1.0.jar" ) );
+
+for ( String path : wanted )
+{
+    if ( jarFile.getEntry( path ) == null )
+    {
+        throw new IllegalStateException( "wanted path is missing: " + path );
+    }
+}
+
+for ( String path : unwanted )
+{
+    if ( jarFile.getEntry( path ) != null )
+    {
+        throw new IllegalStateException( "unwanted path is present: " + path );
+    }
+}
+
+jarFile.close();
--- a/src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
+++ b/src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
@@ -20,6 +20,7 @@ package org.apache.maven.plugins.shade.filter;
  */
 
 import org.apache.maven.artifact.Artifact;
+import org.apache.maven.artifact.DependencyResolutionRequiredException;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.util.IOUtil;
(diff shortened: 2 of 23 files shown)


mkarg:
@rfscholte Review please. :-)

rfscholte:
I see no reason for an extra parameter. If an implementation of a service is provided, all its depending classes should be added, always. So I'd suggest to simplify that part of the code :)

mkarg:
@rfscholte Okay so I will remove the extra parameter.

mkarg:
@rfscholte Removed extra parameter as you proposed, also fixed several bugs. LGTY?

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
```diff
@@ -138,13 +140,19 @@ private void removeServices( final MavenProject project, final Clazzpath cp )
...
                         try ( final BufferedReader bufferedReader =
-                           new BufferedReader( new InputStreamReader( jar.getInputStream( jarEntry ) ) ) )
+                           new BufferedReader( new InputStreamReader( jar.getInputStream( jarEntry ), UTF_8 ) ) )
                         {
                             for ( String className = bufferedReader.readLine(); className != null;
                                 className = bufferedReader.readLine() )
                             {
-                               log.info( className + " was not removed because it is a service" );
-                               removeClass( cp , className );
+                                className = className.split( "#", 2 )[0].trim();
```
rfscholte: I don't understand this. When I remove this, all still works. Please explain.

mkarg: The description of `ServiceClass` says that a line in a service descriptor file may optionally be appended by a hash-separated comment. That line cuts away such optional comments.

rfscholte: Okay, from [ServiceLoader](https://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html)

> This file contains the single line:
> 
>  com.example.impl.StandardCodecs    # Standard codecs

So let's add a test for that too, just ensure nobody removes this piece of code in the future.

mkarg: I will do that, but it will need a while -- have to find a small service JAR on Maven Central that actually contains such a line.

rfscholte: I have a better proposal: let me improve the ITs with the mock repository manager, so we can create such jar ourselves.
Other option: turn the IT into a Maven multimodule project.

mkarg: fine with both, which one do you prefer?

rfscholte: let's go for the latter, I'm still working on other projects and introducing MRM might take some time.

mkarg:
ok I turn the IT into multi-module with self-made service JAR later this week :-)

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
mkarg: ok I turn the IT into multi-module with self-made service JAR later this week :-)

mkarg: Done. :-)

What I do not understand is why I need https://github.com/apache/maven-shade-plugin/pull/17/commits/e4027ae805a903c74a24a9e3baca5069fd42684f. Without those explicit exclusions, maven-rat-plugin asks for a licence header in maven-compiler-plugin's `*.lst` logs!?

rfscholte: looks like the IT is executed in the source-tree; instead it should be moved to target first. But I will take care of that.

mkarg: Great, thanks! :-)

mkarg:
How to go on?

rfscholte:
A little bit of patient, I only have 1 set of hands.

mkarg:
> A little bit of patient, I only have 1 set of hands.

Really? How old school. You should apply for a second set. ;-)

mkarg:
@rfscholte Done with my code change: Only actually needed services (i. e. serviced needed by non-removed classes, and services needed by services needed by services ... to arbitraty depth) are kept. LGTY? :-)

src/it/MSHADE-313/verify.bsh
```diff
@@ -0,0 +1,59 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import java.io.*;
+import java.util.jar.*;
+
+String[] wanted =
```
rfscholte: I'm missing the `META-INF/services/DependencyServiceInterface`, which is quite critical to be verified, right?

mkarg: As minijar only filters `*.class` files, and as [MSHADE-313](#) does not change that strategy, I did not see any good reason to check for the existence this resource file in the outcome (why should it be gone just by adding more class files?). MSHADE-313 does not provide a complete solution for service handling, it just fixes one single bug: prevent deleting needed classes. But as you want to have that check, I just added it right now in 440d6eb0340c7812a8130b3f960739f254e54ea0.

mkarg:
@rfscholte Fixed all you asked for. Please continue review. :-)

rfscholte:
Can you squash these commits?

mkarg:
@rfscholte Squash done. :-)

rfscholte:
Added in https://github.com/apache/maven-shade-plugin/commit/8278c9f28d94f8af6d98655b3db7bd295b9785aa

jira-importer:
Resolve #292

PR #19, opened 2019-04-13
[MSHADE-316] - Configuration option <excludeDefaults>

mkarg:
This PR adds a new configuration option `<excludeDefaults>` (default: `true` for backwards compatibility).

When set to `false`, providing an `<include>` will *not* automatically exclude all other files _anymore_.

This is useful e. g. when minijar'ing a dependency which is loaded by reflection (hence: no class is referenced but just a string is given, so a *specific include* is needed): Without this option, explicitly specifying the reflection-loaded class name would lead to an exclusion of _all_ other files!

Closes [MSHADE-316](https://issues.apache.org/jira/browse/MSHADE-316).

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/filter/SimpleFilter.java
+++ b/src/main/java/org/apache/maven/plugins/shade/filter/SimpleFilter.java
@@ -23,7 +23,10 @@ import org.codehaus.plexus.util.SelectorUtils;
 
 import java.io.File;
 import java.util.HashSet;
+import java.util.Collections;
 import java.util.Set;
+
+import org.apache.maven.plugins.shade.mojo.ArchiveFilter;
 
 /**
  * @author David Blevins
@@ -42,16 +45,42 @@ public class SimpleFilter
 
     private Set<String> excludes;
 
+    private boolean excludeDefaults = true;
+
+    /**
+     * @deprecated As of release 3.2.2, replaced by {@link #SimpleFilter(Set<File>, ArchiveFilter)}
+     * @param jars set of {@link File}s.
+     * @param includes set of includes.
+     * @param excludes set of excludes.
+     */
+    @Deprecated
+    public SimpleFilter( Set<File> jars, Set<String> includes, Set<String> excludes )
+    {
+        this( jars, includes, excludes, true );
+    }
+
+    /**
+     * @param jars set of {@link File}s.
+     * @param archiveFilters set of {@link ArchiveFilter}s.
+     */
+    public SimpleFilter( final Set<File> jars, final ArchiveFilter archiveFilter )
+    {
+        this( jars, archiveFilter.getIncludes(), archiveFilter.getExcludes(), archiveFilter.getExcludeDefaults() );
+    }
+
     /**
      * @param jars set of {@link File}s.
      * @param includes set of includes.
-     * @param excludes set of excludes
+     * @param excludes set of excludes.
+     * @param excludeDefaults whether to exclude default includes once includes are provided explicitly.
      */
-    public SimpleFilter( Set<File> jars, Set<String> includes, Set<String> excludes )
+    private SimpleFilter( final Set<File> jars, final Set<String> includes, final Set<String> excludes,
+      final boolean excludeDefaults )
     {
-        this.jars = ( jars != null ) ? new HashSet<>( jars ) : new HashSet<File>();
+        this.jars = ( jars != null ) ? Collections.<File>unmodifiableSet( jars ) : Collections.<File>emptySet();
         this.includes = normalizePatterns( includes );
         this.excludes = normalizePatterns( excludes );
+        this.excludeDefaults = excludeDefaults;
     }
 
     /** {@inheritDoc} */
@@ -65,7 +94,7 @@ public class SimpleFilter
     {
         String path = normalizePath( classFile );
 
-        return !( isIncluded( path ) && !isExcluded( path ) );
+        return !( ( !excludeDefaults || isIncluded( path ) ) && !isExcluded( path ) );
     }
 
     /**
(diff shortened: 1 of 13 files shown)


mkarg:
@rfscholte Review please. :-)

src/main/java/org/apache/maven/plugins/shade/filter/SimpleFilter.java
```diff
@@ -42,16 +42,19 @@
...
 
+    private boolean excludeDefaults = true;
+
     /**
      * @param jars set of {@link File}s.
      * @param includes set of includes.
      * @param excludes set of excludes
      */
-    public SimpleFilter( Set<File> jars, Set<String> includes, Set<String> excludes )
+    public SimpleFilter( Set<File> jars, Set<String> includes, Set<String> excludes, boolean excludeDefaults )
```
rfscholte: It looks to me we're at a point where we need a second constructor, using the `ArchiveFilter` as argument instead of includes/excludes/excludeDefaults.

rfscholte:
We should update dthe documentation as part of this PR as well

src/main/java/org/apache/maven/plugins/shade/filter/SimpleFilter.java
mkarg: That would be a refactoring, not just a bug fix. Are you sure you want to me to do that in the same PR?

mkarg:
@rfscholte Added documentation for `excludeDefaults` as part of the filter example. Is that sufficient?

src/main/java/org/apache/maven/plugins/shade/filter/SimpleFilter.java
rfscholte: IMHO this is more an improvement than a bugfix. And I'm not sure if others use this SimpleFilter as superclass, so better to keep the original constructor anyway. I don't like the idea that every new field for ArchiveFilter  would translate to an extra constructor, just pass the original instead.

rfscholte:
docs look good enough to me

src/main/java/org/apache/maven/plugins/shade/filter/SimpleFilter.java
mkarg: Implemented additional constructor as you requested, see https://github.com/apache/maven-shade-plugin/pull/19/commits/975615b596d7a99dd5ee9999782aa0b62400a2dd.

mkarg:
All requested changes done. Please continue review. :-)

src/it/MSHADE-316/invoker.properties
```diff
@@ -0,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
rfscholte: I don't see any Java 8  specific code, in which case this file can be removed.

mkarg: You're right, thanks for the pointer! I have removed that file in ec4c056ac338e4129d72fb81fc51a4b3213ef258.

mkarg:
All requested changes done. Please continue review. :-)

mkarg:
@rfscholte Ping ;-)

rfscholte:
I didn't forget this PR, you just have to wait in line.

mkarg:
@rfscholte I really don't wanted to bother you. But see, waiting is stealing drive. It is no critic at all, but just a proposal: Maybe you want me ask other committers instead? I just need to know who is the one to ask for each plugin project.

mkarg:
@rfscholte Thanks, Karl-Heinz, for chiming in. But I cannot merge that. I am only a contributor.

khmarbaise:
Hi @mkarg I know... But If @rfscholte  has no objections I can do this....

khmarbaise:
Hi @mkarg  can you squash your commits into a single commit please....

mkarg:
@khmarbaise Squashed into single commit. BTW, I wonder why you asked me to do that, as Github can automatically squash while merging a PR.

mkarg:
@khmarbaise Six days ago I sqashed the commits as you requested. All requested changes are done since two weeks. I understand that everybody is busy (so am I), but to get new contributors into the boat we need to find a solution to shorten the delays by at least one week. Shall I trigger other committers, e. g. the most active ones according to Github Insights? At Eclipse JAX-RS, where I was nominated for project leadership, we (volunteer committers) have organized to review contributions within one week. So we could help with expertise in organization and onboarding of contributors if you want (just like we helped by asking you for a JavaLand workshop).

olamy:
@rfscholte why are you not using merging pr here? at least this should be close because it's confusing as it has been done in an other way...

rfscholte:
Fixed with https://github.com/apache/maven-shade-plugin/commit/2f6663c725d5b7fd42a177d530eb3577f370f323

mkarg:
@rfscholte Thank you for fixing that small adjustment and for your kind review. But may I ask you, next time when you merge my work, to keep me as the original author? Due to your (apparently uncommon) way of using git, my work now is listed as your work in the git history. While this might not bother most people, actually I have more fun in volunteering when I see *my* photo beside *my* work. If this was by incident, I would kindly ask you to amend that commit by the correct author declaration. Thanks.

rfscholte:
AFAIK that means I must accept the PR via Github and that you must apply all my changes I want as well. In the end I want only 1 working commit, which wasn't the issue in this case.

mkarg:
@rfscholte This is not true. You can work as you did. You just need to *fix* the author line after you're done (before you push) using `git commit --amend --author="Markus KARG <markus@headcrashing.eu>"`. That's all. You can do it even *now*, and `git push -f` after that. :-)

jira-importer:
Resolve #317

PR #28, opened 2019-09-28
[MSHADE-298] GroovyResourceTransformer fix

mplain:
The current implementation of GroovyResourceTransformer does not work with the latest versions of Groovy. As of Groovy 2.5.x, the ExtensionModule files are located in META-INF/groovy, not in META-INF/services.

Old path: META-INF/services/org.codehaus.groovy.runtime.ExtensionModule
New path: META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule

See open issue https://issues.apache.org/jira/browse/MSHADE-298

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/resource/GroovyResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/GroovyResourceTransformer.java
@@ -39,7 +39,7 @@ public class GroovyResourceTransformer
     implements ResourceTransformer
 {
 
-    static final String EXT_MODULE_NAME = "META-INF/services/org.codehaus.groovy.runtime.ExtensionModule";
+    static final String EXT_MODULE_NAME = "META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule";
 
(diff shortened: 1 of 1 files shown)


rmannibucau:
Hi

Shouldnt it be another transformer for backward compatibility?

rfscholte:
I agree with @rmannibucau . One other option is to verify the groovy-version. If it is before 2.5.x, stick to `services`, otherwise switch to `groovy`.
You can get the version by reading the `META-INF/maven/GROUPID/ARTIFACTID/pom.properties`

rfscholte:
And please provide integration-tests in this case for both groovy versions, have a look at `src/it` for inspiration

rfscholte:
Superceded by https://github.com/apache/maven-shade-plugin/commit/799e02abdf936d6f83026403a5ed64e70ad920ef which will keep the Transformer backwords compatible.

mplain:
Thank you! :)

jira-importer:
Resolve #302

jira-importer:
Resolve #302

PR #29, opened 2019-10-11
upgrade to asm 7.2

sullis:

       pulling in other changes.
       commit message.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
     <mavenVersion>3.0</mavenVersion>
     <javaVersion>7</javaVersion>
     <currentVersion>${project.version}</currentVersion>
-    <asmVersion>7.0</asmVersion>
+    <asmVersion>7.2</asmVersion>
   </properties>
 
   <contributors>


Tibor17:
Here is the release notes [support for J13 and J14](https://asm.ow2.io/versions.html).
@rfscholte This version `7.2` should be set in https://github.com/codehaus-plexus/plexus-languages/blob/master/plexus-java/pom.xml#L70 as well and then we can release plugins which are dependent on `plexus-java`.

eolivelli:
What is the better approach?
Wait for a plexus release or patch this plugin?
I see pros and cons in both of the two ways

Tibor17:
@eolivelli Let's merge this fix and release this plugin. The Plexus Java is used in Surefire but it is Codehaus, not Maven gid.

Tibor17:
@sullis pls create Jira ticket and rewrite the commit message, thx.

rfscholte:
> Do you agree with updating plexus-java and cutting a new version 1.0.4?

ASM is a direct dependency of maven-shade plugin, so this PR is correct. No need to worry about plexus-java, the ASM code won't be used.

Tibor17:
It's simple. I mentioned plexus-java because surefire is dependent on it and thus transitive dep ASM. So we have a common problem, updates due to new J13 and J14.

PR #30, opened 2019-11-01
[MSHADE-284] - Shaded test JARs are always empty

peterdemaeyer:
This PR fixes [MSHADE-284](https://issues.apache.org/jira/browse/MSHADE-284), which causes the test JAR to be always empty. The shadeTestJar feature was broken from day one, has never ever worked, until now.
- IT added (all the new files in this PR are for the IT).
- Code fix in ShadeMojo.java.
- Test fails before fix, succeeds after.
- "mvn clean verify" succeeds.
- "mvn -Prun-its clean verify" succeeds.
 - [x] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

[After all these years, once this PR gets merged, shadeTestJar will _finally_ work like it was supposed to, and people can stop using the assembly plugin as a workaround.](https://stackoverflow.com/questions/5149130/how-can-i-configure-the-maven-shade-plugin-to-include-test-code-in-my-jar)

### DIFF
--- a/src/it/MSHADE-284_shadeTestJar/api/pom.xml
+++ b/src/it/MSHADE-284_shadeTestJar/api/pom.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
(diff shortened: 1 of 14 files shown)


struberg:
Thanks for the PR, Peter!

jira-importer:
Resolve #291

PR #35, opened 2020-01-03
[MSHADE-311] - Don't add exclusions for provided dep

famod:
Since the Exclusion logic in `ShadeMojo.updateExcludesInDeps(MavenProject, List<Dependency>, List<Dependency>)` was introduced for `promoteTransitiveDependencies=true` (see https://github.com/apache/maven-shade-plugin/commit/13c6097fffaa85d1ea40752cef732d6548e43e2c and [MSHADE-31](https://issues.apache.org/jira/browse/MSHADE-31)) which doesn't affect `provided` dependencies anyway, a simple scope check was sufficient to fix this problem.

       pulling in other changes.
       commit message.


famod:
Btw: I find it very unfortunate that back then MSHADE-31 did _not_ introduce a test.
So my new integration test seems to be the first test actually testing the dependency reduced POM for the presence/absence of `exclusions` (but only for a very specific scenario).

In case `prodvided` scope dependencies are ever fully adressed by this plugin (which would make sense by the way, see also [MSHADE-89](http://jira.codehaus.org/browse/MSHADE-89) and [MSHADE-181](http://jira.codehaus.org/browse/MSHADE-181)), the entire `exclusions` handling should be re-implemented. The current approach might have been influenced by older Maven versions, but from a recent POV: Why isn't the Plugin just checking the original exclusions of the original pom? Instead is is (IMHO) trying to "guess" what _might_ have been an exclusion before all the shading.

khmarbaise:
Let's see what our CI says https://builds.apache.org/view/M-R/view/Maven/job/maven-box/job/maven-shade-plugin/job/famod-MSHADE-311-dep-red-pom-excludes/

khmarbaise:
So CI is fine..will be merged into master. Thanks for your contribution.

khmarbaise:
Has been merged into master.

famod:
Thank you very much, @khmarbaise!

jira-importer:
Resolve #312

PR #38, opened 2020-02-06
[MSHADE-350] enable to rewrite the manifest

rmannibucau:
Goal is to be able to use relocations in manifest imports/exports/capabilities as we do for service transformers - otherwise relocations break OSGi integration.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/resource/ManifestResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/ManifestResourceTransformer.java
@@ -51,6 +51,8 @@ public class ManifestResourceTransformer
     // Fields
     private boolean manifestDiscovered;
 
+    private boolean skipRelocators;
+
     private Manifest manifest;
 
     public boolean canTransformResource( String resource )
@@ -72,7 +74,37 @@ public class ManifestResourceTransformer
         if ( !manifestDiscovered )
         {
             manifest = new Manifest( is );
+
+            if ( relocators != null && !relocators.isEmpty() )
+            {
+                rewriteManifest( relocators );
+            }
+
             manifestDiscovered = true;
+        }
+    }
+
+    private void rewriteManifest( List<Relocator> relocators )
+    {
+        if ( skipRelocators )
+        {
+            return;
+        }
+        for ( final Map.Entry<Object, Object> entry : manifest.getMainAttributes().entrySet() )
+        {
+            final Object value = entry.getValue();
+            if ( value instanceof  String )
+            {
+                String relContent = String.valueOf( value );
+                for ( final Relocator relocator : relocators )
+                {
+                    relContent = relocator.applyToSourceContent( relContent );
+                }
+                if ( !relContent.equals( value ) )
+                {
+                    entry.setValue( relContent );
+                }
+            }
         }
     }
 
--- a/src/main/java/org/apache/maven/plugins/shade/resource/ManifestResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/ManifestResourceTransformer.java
@@ -46,7 +46,9 @@ public class ManifestResourceTransformer
     // Configuration
     private String mainClass;
 
-    private boolean skipRelocators;
+    private boolean useDefaultLocators;
+
+    private List<Relocator> relocators;
 
     private Map<String, Object> manifestEntries;
 
@@ -75,9 +77,13 @@ public class ManifestResourceTransformer
         {
             manifest = new Manifest( is );
 
-            if ( relocators != null && !relocators.isEmpty() )
+            if ( relocators != null && !relocators.isEmpty() && useDefaultLocators )
             {
                 rewriteManifest( relocators );
+            }
+            if ( this.relocators != null && !this.relocators.isEmpty() )
+            {
+                rewriteManifest( this.relocators );
             }
 
             manifestDiscovered = true;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/maven/plugins/shade/resource/ManifestResourceTransformer.java
```diff
@@ -51,6 +51,8 @@
     // Fields
     private boolean manifestDiscovered;
 
+    private boolean skipRelocators;
```
eolivelli: it looks like you are never setting a value

rmannibucau: it is set in the pom.xml and is an user config. Guess I forgot a setter, will add it.

src/main/java/org/apache/maven/plugins/shade/resource/ManifestResourceTransformer.java
```diff
@@ -46,6 +46,8 @@
     // Configuration
     private String mainClass;
 
+    private boolean skipRelocators;
```
rfscholte: Do we really need this skip parameter? If the manifest file should be transformed and there are relocated classes, just apply them. I see no reason to pollute this class with a skip parameter.

rmannibucau: Agree but the algorithm being a bit fragile until we use a full osgi header parsers/rewriter with exception algorithms for the general case, I'd prefer a parameter to disable it for not osgi case. I can enhance the parsing but thought it is not the core of shade plugin so is kot worse that logic. If wrong, happy to revise this.
Not perfect state but secured - would be sad to break a project for that new behavior IMHO.
Wdyt?

rfscholte: Is there a way to discover if it is a OSGi project, e.g. via the packaging?

rmannibucau: Yes and no, they will have Bundle-Version entry in the manifest but most of the modules will be potential bundles but more often used in se/ee. For ex, the test i wrote used an openwebbeans output. It is used in se context, tomee, meecrowave (all not osgi) but also aries-cdi (osgi) or as plain osgi bundle.
This is why i introduced this boolean.

rfscholte: Looking at the code, it should be possible with a dedicated relocator instead of the SimpleRelocator. In this case you'll need a more delicate way than the brute search and replace all.

rmannibucau: Hmm, sounds like an option but current code works well for jakarta relocation too so I'm mixed to make the conf more complex/verbose for end user.
Is having the boolean AND dedicated relocators ok for you? Boolean could be reversed and named "inheritDefaultRelocators".
Wdyt?

rmannibucau: FYI, I updated the code to reflect the proposal

rfscholte: Once a skip parameter is introduced, it is very hard to get rid of it, so I'll try to block it when possible(it is not a common practice as many users use as argument, it is just a lazy hack). The enduser shouldn't notice the difference. I'm also working on the maven-shade-plugin, so I'll give it a try as well.

rfscholte:
My counter proposal: https://github.com/apache/maven-shade-plugin/commit/a3ca8bb7297e4e8db5ccecd89c1adf64cd6ce506 (reusing your testcase)

rmannibucau:
@rfscholte means just supporting a limited set of manifest entries - standard ones I guess? i dont think it will be sufficient and generic enough, would be sad to have to write a custom transformer just for that IMHO, no?

rfscholte:
The Manifest file is tricky: you can't predict if an attribute contains a class or not. So the safest way is to be explicit. Ideally I would even analyze current entries to ensure only the package-parts are patched.
If you want it more flexible, then you should add the option to select the attributes to patch, but I wonder if we're really helping users with such flexibility if we already know which attributes contain classes or packages.

rmannibucau:
@rfscholte fact is we can't know (at job i use custom entries with classes as well). Parsing is an option for known entries, not for custom ones (what if i put json ;)). Also for prefixes, reusing the global relocators are the simplest option so having to toggle to inherit from them make sense for me even if false by default.

rfscholte:
We shouldn't try to make things bigger than necessary. You provided a test (brilliant) with known OSGi attributes. After the patch the test succeeds.
Does it require extra effort and customization from the user? No.
Can we extend it when required? Sure, either by increasing the list of known attributes or giving the ability to specify this list. But we should only do the latter if we can't solve it with known attributes anymore.

If you also want to support custom attributes, you should add a test based on a realworld example. (you could decide to put that in a new ticket)

rmannibucau:
@rfscholte the test enables the defaultLocators usage (useDefaultLocators.set(transformer, true);), if I want it to pass without this toggle, I need to copy the global relocators to the transformers - which is what I tried to avoid. Concretely I guess X-CompanyCensured-ClassesConfigurationInJson can't be "known" but falls into the category of things to rewrite. If you propose a List<String> attributesToRewrite, or includes/exclude keys, I'm fine with that as well.

rfscholte:
I've updated my proposal to support additional atttributes

rmannibucau:
Ok for me Robert. Thanks.

rfscholte:
superceded by https://github.com/apache/maven-shade-plugin/commit/0aa56da230bd0c86b8ffe7a8e6405b11dce4309a

jira-importer:
Resolve #357

PR #42, opened 2020-04-11
[MSHADE-359] update guava

elharo:
@michael-o  These are the easy and relatively safe changes to remove the dependency on Guava and stick to Apache classes. There's one more usage of Multimap in core code that's going to need more careful work before we can rip it out completely.


jira-importer:
Resolve #367

PR #73, opened 2020-10-07
[MSHADE-378] Shade plugin changes the compression level of nested jar…

jenrryyou:
Shade plugin loses the meta information of orignal jar entries when copying nested jar, which make these nested jar can't be loaded by JVM。This PR is trying to fix it by modified the copy logic for JarEntry.

The jira ticket link: https://issues.apache.org/jira/projects/MSHADE/issues/MSHADE-378?filter=allopenissues

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
+++ b/src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
@@ -20,15 +20,14 @@ package org.apache.maven.plugins.shade;
  */
 
 import java.io.BufferedOutputStream;
-import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
-import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
+import java.io.PushbackInputStream;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


rmannibucau:
Hi @jenrryyou ,

the PR looks good but is it possible to add a test about this case (thinking out loud DefaultShaderTest can be a base and required to generate a fake jar with nested jar using TemporaryFolder as base).

Impl side note: a PushbackInputStream can probably avoid a custom inputstream impl and still enable to test the zip header.

jenrryyou:
Hi @rmannibucau ,
thanks for your review and comment，I will add a test for this case. As for using PushbackInputStream, it seems like a better 
idea and I will try it also.

jenrryyou:
Hi @rmannibucau ,
A test is added for this PR to verify that shade plugin wouldn't change the compression level of nested jar.  By the way,  ZipHeaderPeekInputStream is refined and  based on PushbackInputStream now.

rmannibucau (approved):

jenrryyou:
[MSHADE-378] Shade plugin changes the compression level of nested jar entries, see more at this [jira ticket link](https://issues.apache.org/jira/projects/MSHADE/issues/MSHADE-378?filter=allopenissues)

jenrryyou:
Hi @rmannibucau ,
Since I have no write access for this repo，I was wondering if you could merge this PR or somebody else would do?

jira-importer:
Resolve #684

jira-importer:
Resolve #684

PR #92, opened 2021-05-19
[MSHADE-390] Sisu Index transformer

cstamas:
https://issues.apache.org/jira/browse/MSHADE-390

Adds SISU index transformer to existing ones + IT.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
@@ -19,14 +19,13 @@ package org.apache.maven.plugins.shade.resource;
  * under the License.
  */
 
-import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.StringReader;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.Scanner;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
 
@@ -48,8 +47,6 @@ public class SisuIndexResourceTransformer
 
     private ServiceStream serviceStream;
 
-    private List<Relocator> relocators;
-
     private long time = Long.MIN_VALUE;
 
     @Override
@@ -70,12 +67,10 @@ public class SisuIndexResourceTransformer
         }
 
         final String content = IOUtils.toString( is, StandardCharsets.UTF_8 );
-        StringReader reader = new StringReader( content );
-        BufferedReader lineReader = new BufferedReader( reader );
-        String line;
-        while ( ( line = lineReader.readLine() ) != null )
+        Scanner scanner = new Scanner( content );
+        while ( scanner.hasNextLine() )
         {
-            String relContent = line;
+            String relContent = scanner.nextLine();
             for ( Relocator relocator : relocators )
             {
                 if ( relocator.canRelocateClass( relContent ) )
@@ -83,12 +78,8 @@ public class SisuIndexResourceTransformer
                     relContent = relocator.applyToSourceContent( relContent );
                 }
             }
-            serviceStream.append( relContent + NEWLINE );
-        }
-
-        if ( this.relocators == null )
-        {
-            this.relocators = relocators;
+            serviceStream.append( relContent );
+            serviceStream.append( NEWLINE );
         }
 
         if ( time > this.time )
@@ -129,7 +120,6 @@ public class SisuIndexResourceTransformer
         {
             byte[] contentBytes = content.getBytes( StandardCharsets.UTF_8 );
             this.write( contentBytes );
-            this.write( NEWLINE.getBytes( StandardCharsets.UTF_8 ) );
         }
 
         public InputStream toInputStream()
--- a/src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
@@ -42,8 +42,6 @@ public class SisuIndexResourceTransformer
 {
     private static final String SISU_INDEX_PATH = "META-INF/sisu/javax.inject.Named";
 
-    private static final String NEWLINE = "\n";
-
     private final ArrayList<String> indexEntries = new ArrayList<>();
 
     private long time = Long.MIN_VALUE;
@@ -93,7 +91,7 @@ public class SisuIndexResourceTransformer
         JarEntry jarEntry = new JarEntry( SISU_INDEX_PATH );
         jarEntry.setTime( time );
         jos.putNextEntry( jarEntry );
-        IOUtils.writeLines( indexEntries, NEWLINE, jos, StandardCharsets.UTF_8 );
+        IOUtils.writeLines( indexEntries, "\n", jos, StandardCharsets.UTF_8 );
         jos.flush();
         indexEntries.clear();
    }
--- a/src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
@@ -1,0 +1,162 @@
+package org.apache.maven.plugins.shade.resource;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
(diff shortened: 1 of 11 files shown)


src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,162 @@
...
+
+        JarEntry jarEntry = new JarEntry( SISU_INDEX_PATH );
+        jarEntry.setTime( time );
+        jos.putNextEntry( jarEntry );
+
+
+        // read the content of service file for candidate classes for relocation.
+        // Specification requires that this file is encoded in UTF-8.
+        Writer writer = new OutputStreamWriter( jos, StandardCharsets.UTF_8 );
+        InputStreamReader streamReader = new InputStreamReader( data.toInputStream() );
```
elharo, line 121: needs character set, probably UTF-8

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,162 @@
...
+        BufferedReader reader = new BufferedReader( streamReader );
+        String className;
+
+        while ( ( className = reader.readLine() ) != null )
```
elharo, line 125: commons IO probably has this logic for copying streams already

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,162 @@
...
+        {
+            writer.write( className );
+            writer.write( System.lineSeparator() );
+            writer.flush();
+        }
+
+        reader.close();
```
elharo, line 132: use try finally

cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+        {
+            super( 1024 );
+        }
+
+        public void append( String content )
+            throws IOException
+        {
+            byte[] contentBytes = content.getBytes( StandardCharsets.UTF_8 );
+            this.write( contentBytes );
+            this.write( NEWLINE.getBytes( StandardCharsets.UTF_8 ) );
```
cstamas, line 132: This double NL

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+public class SisuIndexResourceTransformer
+    extends AbstractCompatibilityTransformer
+{
+    private static final String SISU_INDEX_PATH = "META-INF/sisu/javax.inject.Named";
+
+    private static final String NEWLINE = "\n";
+
+    private ServiceStream serviceStream;
+
+    private List<Relocator> relocators;
```
cstamas, line 51: Drop this

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.jar.JarEntry;
+import java.util.jar.JarOutputStream;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.maven.plugins.shade.relocation.Relocator;
+
+/**
+ * Resources transformer that relocates classes in {@code META-INF/sisu/javax.inject.Named} and appends resources
```
elharo, line 37: Resources --> Resource

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+                                 final List<Relocator> relocators,
+                                 long time ) throws IOException
+    {
+        if ( serviceStream == null )
+        {
+            serviceStream = new ServiceStream();
+        }
+
+        final String content = IOUtils.toString( is, StandardCharsets.UTF_8 );
+        StringReader reader = new StringReader( content );
```
elharo, line 73: This seems unnecessary. You can read lines directly from the stream without storing it in a string first.

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: using scanner now

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+                                 final List<Relocator> relocators,
+                                 long time ) throws IOException
+    {
+        if ( serviceStream == null )
+        {
+            serviceStream = new ServiceStream();
+        }
+
+        final String content = IOUtils.toString( is, StandardCharsets.UTF_8 );
+        Scanner scanner = new Scanner( content );
```
elharo, line 70: create the scanner from the input stream and charset, no need to store in a string first

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    public void modifyOutputStream( final JarOutputStream jos )
+        throws IOException
+    {
+        JarEntry jarEntry = new JarEntry( SISU_INDEX_PATH );
+        jarEntry.setTime( time );
+        jos.putNextEntry( jarEntry );
+        IOUtils.copy( serviceStream.toInputStream(), jos );
```
elharo, line 104: looks like a possible nullpointerexception here since serviceStream can be null

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+        jos.flush();
+        serviceStream.reset();
+   }
+
+    static class ServiceStream
```
elharo, line 109: I suspect there's a simpler way to accomplish this, likely with code in the JDK or perhaps commons.io. Yes, I think you can forget about copying everything into streams. Just build a string when you transform and write that into the jar.

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+        {
+            String relContent = scanner.nextLine();
+            for ( Relocator relocator : relocators )
+            {
+                if ( relocator.canRelocateClass( relContent ) )
+                {
+                    relContent = relocator.applyToSourceContent( relContent );
+                }
+            }
+            serviceStream.append( relContent );
```
elharo, line 81: I don't think you need these streams here. Just build a String and work with that.

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: Theoretically yes, but transformer is invoked like this (in DefaultShader):
```java
                if ( transformer.hasTransformedResource() )
                {
                    transformer.modifyOutputStream( out );
                }
```

And if hasTransformedResource is true, that means serviceStream is not null. Doing check here seems superfluous to me, but I'd agree to throw something here if method invoked BUT serviceStream is null.

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: agreed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: yup, much-o-simpler indeed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: fixed

src/main/java/org/apache/maven/plugins/shade/resource/SisuIndexResourceTransformer.java
cstamas: this is fluke since last changes (not possible to end up with NPE)

elharo (approved):

cstamas:
merged as e6902dd34a917014864162ca05d06fee14f4e4c5

jira-importer:
Resolve #396

jira-importer:
Resolve #396

jira-importer:
Resolve #396

PR #93, opened 2021-05-20
Improve and simplify ServicesResourceTransformer

cstamas:
Payback time: As ServicesResourceTransformer did "inspire"
SisuIndexResourceTransformer, and the latter with elharo
help came out rather nice, I think it is "payback
time" and improve similarly the original inspiration as well.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/pom/MavenJDOMWriter.java
+++ b/src/main/java/org/apache/maven/plugins/shade/pom/MavenJDOMWriter.java
@@ -29,6 +29,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.maven.model.ActivationFile;
 import org.apache.maven.model.ActivationOS;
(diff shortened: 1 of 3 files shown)


elharo (approved):

cstamas:
Bah, iirc the it enforces extra newline between merged segments, will fix

cstamas:
So, as I assumed: it is OS newline vs Java `\n`. But, the good news is that ONLY TWO classes in this project are affected by it (uses `System.getProperty( "line.separator" )`):
* the MavenJDOMWriter
* and the ServiceResourceTransformerTest UT that failed on CI. This is wrong IMO, as https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html tells about file encoding, but IMO line ending of `\n` for it (on any plaf) is acceptable.

All the rest of code use `\n`.

UT is really unimportant (will align it with this PR), but unsure why MavenJDOMWriter uses OS newline, when IMHO it should be either using "edited POM" lineending or universal `\n`.

@elharo @michael-o any opinion?

cstamas:
@hboutemy expecting your response as well, as EOL thing affects reproducible builds.... (as same git hash on win or non-win produces different "edited" POM)

elharo (commented):
Almost all code should be using a consistent line separator regardless of platform. That said, I'm not surprised you found some code where this was done wrong. I've found quite a bit of code scattered throughout the project that does not use platform independent line separators.

hboutemy:
platform dependent newline is not an issue, it's normal, even if sometimes it adds some complexity: it does not block Reproducible Builds, see https://github.com/jvm-repo-rebuild/reproducible-central where I reproduced many releases done on Windows (even if I build with a Linux Docker image)

in case of shade transformer, thinking deeper at it, you are not forced to keep original newline: you can rewrite the file if you need it (and by "need", it can mean "makes your life easier"), the transformer perfectly knows the format of what it transforms, what is important is that:
1. the output is valid
2. the output is Reproducible, that is 2 runs give the same output (no randomness added, newline not being seen as randomness but reproducibility environment requirement = accepted constraint)

so choose the strategy you wish on newline for this file: it's a local personal choice that won't affect other parts of the build where platform newline is used.
IMHO only one choice is to be avoided: mix different newlines in one file. Even that one is a question of taste, because it does not break the 2 previous key requirements = valid + reproducible

cstamas:
merged as c18d3d3e36f4d37aba98a18983bd4e56eba79327

PR #88, opened 2021-04-04
[MSHADE-252] - Make source shading respect package/class name excludes

kriegaex:
**1. Make source shading respect package/class name excludes**

Fixes MSHADE-252.

Until now, there was a naive global search + replace for mapping `<pattern>` to `<shadedPattern>`, even if the corresponding packages are to be ignored. This works for binary JARs, but was completely broken for source JARs.

See also my related [comment in MSHADE-252](https://issues.apache.org/jira/browse/MSHADE-252?focusedCommentId=17314377&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17314377).

With this fix, on the one hand string replacement in source files works for both "dotty" strings like `my.package.name.Foo` and "slashy" ones like `/my/package/name/resource.properties`. On the other hand, source files are now being parsed `<pattern>` match by match and for each match the excludes list will be checked and replacements only performed if the corresponding matched package name is not on the excludes list. This is of course slower than just a raw `replaceAll(..)`, but it produces correct results.

---

**2. Make relocation ex-/include matching more lenient, accepting `.**` too**

Actually, class patterns are supposed to should just use `foo.bar.*` ending with a single asterisk, but some users mistake them for path patterns like `my/path/**`, so let us be a bit more lenient here.

See also my related [comment in MSHADE-252](https://issues.apache.org/jira/browse/MSHADE-252?focusedCommentId=17314393&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17314393).

---

I have just sent my **Contributor License Agreement** to the Apache secretary's e-mail address. I hope is shall be processed quickly.

---

Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MSHADE) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MSHADE-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MSHADE-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
[...]

PR #95, opened 2021-05-21
[MSHADE-391] Do not write modified class files for no-op relocations

kriegaex:
Also, rename method `DefaultShader.shadeSingleJar` to `shadeJarEntry` in order to reflect what it actually does: It does **not** shade a JAR but a JAR entry, i.e. not a "single JAR" but rather a single file.

---

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
+++ b/src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
@@ -111,7 +111,7 @@ public class DefaultShader
             }
         }
 
-        final DefaultRelocators relocators = new DefaultRelocators( shadeRequest.getRelocators() );
+        final DefaultPackageMapper packageMapper = new DefaultPackageMapper( shadeRequest.getRelocators() );
 
         // noinspection ResultOfMethodCallIgnored
         shadeRequest.getUberJar().getParentFile().mkdirs();
@@ -125,7 +125,7 @@ public class DefaultShader
             MultiValuedMap<String, File> duplicates = new HashSetValuedHashMap<>( 10000, 3 );
             // CHECKSTYLE_ON: MagicNumber
 
-            shadeJars( shadeRequest, resources, transformers, out, duplicates, relocators );
+            shadeJars( shadeRequest, resources, transformers, out, duplicates, packageMapper );
 
             // CHECKSTYLE_OFF: MagicNumber
             MultiValuedMap<Collection<File>, String> overlapping = new HashSetValuedHashMap<>( 20, 15 );
@@ -224,7 +224,7 @@ public class DefaultShader
 
     private void shadeJars( ShadeRequest shadeRequest, Set<String> resources, List<ResourceTransformer> transformers,
                             JarOutputStream jos, MultiValuedMap<String, File> duplicates,
-                            DefaultRelocators relocators )
+                            DefaultPackageMapper packageMapper )
         throws IOException
     {
         for ( File jar : shadeRequest.getJars() )
@@ -266,7 +266,7 @@ public class DefaultShader
 
                     try
                     {
-                        shadeJarEntry( shadeRequest, resources, transformers, relocators, jos, duplicates, jar,
+                        shadeJarEntry( shadeRequest, resources, transformers, packageMapper, jos, duplicates, jar,
                                         jarFile, entry, name );
                     }
                     catch ( Exception e )
@@ -281,14 +281,14 @@ public class DefaultShader
     }
 
     private void shadeJarEntry( ShadeRequest shadeRequest, Set<String> resources,
-                                 List<ResourceTransformer> transformers, DefaultRelocators relocators,
+                                 List<ResourceTransformer> transformers, DefaultPackageMapper packageMapper,
                                  JarOutputStream jos, MultiValuedMap<String, File> duplicates, File jar,
                                  JarFile jarFile, JarEntry entry, String name )
         throws IOException, MojoExecutionException
     {
         try ( InputStream in = jarFile.getInputStream( entry ) )
         {
-            String mappedName = relocators.map( name, false );
+            String mappedName = packageMapper.map( name, true, false );
 
             int idx = mappedName.lastIndexOf( '/' );
             if ( idx != -1 )
@@ -304,7 +304,7 @@ public class DefaultShader
             duplicates.put( name, jar );
             if ( name.endsWith( ".class" ) )
             {
-                addRemappedClass( jos, jar, name, entry.getTime(), in, relocators );
+                addRemappedClass( jos, jar, name, entry.getTime(), in, packageMapper );
             }
             else if ( shadeRequest.isShadeSourcesContent() && name.endsWith( ".java" ) )
             {
@@ -515,10 +515,10 @@ public class DefaultShader
     }
 
     private void addRemappedClass( JarOutputStream jos, File jar, String name,
-                                   long time, InputStream is, DefaultRelocators relocators )
+                                   long time, InputStream is, DefaultPackageMapper packageMapper )
         throws IOException, MojoExecutionException
     {
-        if ( relocators.relocators.isEmpty() )
+        if ( packageMapper.relocators.isEmpty() )
         {
             try
             {
@@ -550,7 +550,7 @@ public class DefaultShader
         ClassWriter cw = new ClassWriter( 0 );
 
         final String pkg = name.substring( 0, name.lastIndexOf( '/' ) + 1 );
-        final ShadeClassRemapper cv = new ShadeClassRemapper( cw, pkg, relocators );
+        final ShadeClassRemapper cv = new ShadeClassRemapper( cw, pkg, packageMapper );
 
         try
         {
@@ -575,7 +575,7 @@ public class DefaultShader
         }
 
         // Need to take the .class off for remapping evaluation
-        String mappedName = relocators.map( name.substring( 0, name.indexOf( '.' ) ), false );
+        String mappedName = packageMapper.map( name.substring( 0, name.indexOf( '.' ) ), true, false );
 
         try
         {
@@ -687,52 +687,59 @@ public class DefaultShader
         }
     }
 
-    // limit at the minimum the relocating API we use to enable to add transversal features around it
-    // also use an interface to overpass the ASM API limitations (extends ClassRemapper + impl remapper)
-    private interface Relocators
+    private interface PackageMapper
     {
-        String map( String name, boolean clazz );
+        /**
+         * Map an entity name according to the mapping rules known to this package mapper
+         * 
+         * @param entityName entity name to be mapped
+         * @param mapPaths map "slashy" names like paths or internal Java class names, e.g. {@code com/acme/Foo}?
+         * @param mapPackages  map "dotty" names like qualified Java class or package names, e.g. {@code com.acme.Foo}?
+         * @return mapped entity name, e.g. {@code org/apache/acme/Foo} or {@code org.apache.acme.Foo}
+         */
+        String map( String entityName, boolean mapPaths, boolean mapPackages );
     }
 
-    // our relocators facade
-    private static class DefaultRelocators implements Relocators
+    /**
+     * A package mapper based on a list of {@link Relocator}s
+     */
+    private static class DefaultPackageMapper implements PackageMapper
     {
         private static final Pattern CLASS_PATTERN = Pattern.compile( "(\\[*)?L(.+);" );
 
         private final List<Relocator> relocators;
 
-        private DefaultRelocators( final List<Relocator> relocators )
+        private DefaultPackageMapper( final List<Relocator> relocators )
         {
             this.relocators = relocators;
         }
 
         @Override
-        public String map( String name, final boolean relocateClass )
+        public String map( String entityName, boolean mapPaths, final boolean mapPackages )
         {
-            final String originalName = name;
-            String value = originalName;
+            String value = entityName;
 
             String prefix = "";
             String suffix = "";
 
-            Matcher m = CLASS_PATTERN.matcher( name );
+            Matcher m = CLASS_PATTERN.matcher( entityName );
             if ( m.matches() )
             {
                 prefix = m.group( 1 ) + "L";
                 suffix = ";";
-                name = m.group( 2 );
+                entityName = m.group( 2 );
             }
 
             for ( Relocator r : relocators )
             {
-                if ( relocateClass && r.canRelocateClass( name ) )
+                if ( mapPackages && r.canRelocateClass( entityName ) )
                 {
-                    value = prefix + r.relocateClass( name ) + suffix;
+                    value = prefix + r.relocateClass( entityName ) + suffix;
                     break;
                 }
-                else if ( r.canRelocatePath( name ) )
+                else if ( mapPaths && r.canRelocatePath( entityName ) )
                 {
-                    value = prefix + r.relocatePath( name ) + suffix;
+                    value = prefix + r.relocatePath( entityName ) + suffix;
                     break;
                 }
             }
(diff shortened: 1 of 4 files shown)


eolivelli (commented):
great work
thank you very much.

I wonder if you could try to add a test case in order to prevent regressions in the future

src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
```diff
@@ -694,6 +702,11 @@ private void addResource( Set<String> resources, JarOutputStream jos, String nam
 
         List<Relocator> relocators;
 
+        // Use thread-local, just in case 'map*' calls are ever done concurrently. Make sure that the using class
+        // initialises this value according to its needs, usually setting the value to false per file before starting
+        // relocation.
+        ThreadLocal<Boolean> wasRelocated = new ThreadLocal<>();
```
eolivelli, line 708: usually ThreadLocals are declared a "static final" members

rmannibucau: guess there is no need of any threadlocal, the boolean can be stored in a custom ClassRemapper, sounds simpler/less hacky in terms of code to me, wdyt?

kriegaex: @eolivelli, I can make the TL final, no problem.

@rmannibucau, the plugin already uses one `ClassRemapper` per file (i.e. per JAR entry, not per JAR) in order to take care of source file references in each class file. I think we should not create extra objects (one per file in each source JAR) just so as to store a single value. I am not an ASM buff either, BTW. But if the thread-local is not acceptable, I would change it before the PR does not get merged. I am happy to get quick feedback, after my other PR went unnoticed for so long, even though it provides higher value than this one.

kriegaex:
> I wonder if you could try to add a test case in order to prevent regressions in the future

Such a test would be heuristic, relying on whether ASM happens to restructure certain class files during transformation or not. It would also be an integration test with some real world dependencies and classes using them. Indirectly, we would be testing ASM rather than Shade, but the test is certainly doable. It would be more expensive to create than the actual change, though.

kriegaex:
Please also find the TODO in the diff and provide some feedback as to whether or not I should simplify the mapping method to an unparametrised one used by both callers. If so, I would rather inline it and chain the calls.

src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
kriegaex: Update: I forgot to mention, that the custom `Remapper` called `RelocatorRemapper` is instantiated only once, and I did not want to refactor the whole plugin. So I stuck with the idea of one instance and used the thread-local field.

src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
```diff
@@ -706,45 +719,25 @@ public boolean hasRelocators()
...
-            }
-
-            return super.mapValue( object );
+            return object instanceof String ? map( (String) object, true ) : super.mapValue( object );
         }
 
         public String map( String name )
         {
-            String value = name;
+            // TODO: Before the refactoring of duplicate code to 'private String map(String, boolean)', this method did
```
eolivelli, line 727: I am sorry, I am not sure how to help.

I think that keeping the current behaviour is safer.
we can work on this topic on a follow up change

kriegaex: You see the duplication in the original code before my change, but also the slight difference. That might be an oversight or on purpose, but nobody documented the intent of duplicating with a minimal difference. The tests I ran locally (all unit and integration tests plus the project I am working on) did not show any differences if I just always check for both dotty and slashy package names.

src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
rmannibucau: A thread local is a hack in the sense it is a solution to keep track of a data in a multithreaded application with reference breakage which is not the case here. Here you have 1 instance of classreader/writer per class file which can track the state of the rewritten class (plain boolean).
An example of such thing can be seen in sirona: https://github.com/rmannibucau/sirona/blob/trunk/agent/javaagent/src/main/java/com/github/rmannibucau/sirona/javaagent/SironaClassVisitor.java#L139
So no additional instantiation in object count, no plugin refactoring or so and simpler code path since the state becomes an attribute and not an unrelated threadlocal so overall I don't see any reason to use a quick and dirty option for this.

hope it makes sense

kriegaex: Probably it does, but I do not understand how you would implement it, because the methods which knows that a relocation has taken place is `DefaultShader.RelocatorRemapper.map(String, boolean)`, and that object is only instantiated once like I said. It has no reference to the `ClassRemapper` used locally in `DefaultShader.addRemappedClass()`, but the other way around. If I would pass all the way through to the `map` method, it would not only be ugly but also mean that in the remapper it would have to be kept in a thread-local again, because the (quasi singleton) remapper can be called multiple times.

So if you mean something else and have thought this through, either explain it as clearly as you can, so I can implement it according to your idea, or accept the PR with the thread-local and refactor it by yourself in a later stage. What I am presenting is a substantial improvement. I see no need to make it perfect, just better than before without any new bugs. Let us apply the boyscout rule: leave the camp ground behind a little cleaner than we found it, but let us not stay there longer than the actual trip took.

rmannibucau: Start by creating a ShadeClassRemapper (a visitor) and each time remapper can be call test it was called or not.
You can even just use a boolean in the remapper  since it is single thread and can be simpler at the end but in terms of arch it is the visitor which would hold that logic by decorating remapper usages.

Another thing to test is just to compare the byte[] before and after, guess in several cases it will be the same and it can be a sufficient first impl but the remapper decoration is what makes the code the more straight forward and less hacky IMHO.

Can surely hack it next week if you need help about this.

kriegaex: > Start by creating a ShadeClassRemapper... 

It is after midnight here, maybe I can look into it tomorrow when I can think straight again. Thanks. 

> Another thing to test is just to compare the byte[] before and after

No, that is exactly what is **not** working and what my change is about: to determine that no relocation had taken place, even though the byte code is different due to rewriting with ASM. Those different files, even though I was sure they were unrelated to any relocation, were what triggered my attention and made me discover what was going on here. So that idea is a dead end. Maybe check out and use my code, looking at it as a whole instead of focusing on the thread-local, if you have not done so yet, don't just look at it on GitHub.

rmannibucau: I did review it, except this threadlocal issue which will explode in a few more iterations at our face it looks acceptable to me

kriegaex: I didn't say or mean, you didn't review it. You obviously did. I was not sure if you checked out the branch and watched what it does and compared to what it did before. If you did, fine. Then you know the byte code comparison doesn't cut it. Possibly your other idea does. I never understood this cryptic way of nesting and delegating readers, writers, visitors and remappers in ASM. So this is me being inexperienced in that field or maybe also being not smart enough or not conditioned on this kind of usage pattern. But I do promise to try and understand what you mean tomorrow.

rmannibucau: Well it does if you rewrite it twice which is likely slower - didnt test but being in mem we can be surprised, was my point.
Agree asm API is not that nice when used in visitor but it is needed copying a class so let's keep it for this iteration maybe.
Can help with code on monday (too late for this week ;))

kriegaex: Why would I rewrite it twice, other than in a test? In a test, I could rewrite without relocation, then again with, hoping that that way a comparison would be easier. But we are talking about the production code here.

kriegaex: I made the `ThreadLocal` final, as requested. I decided not to refactor to create more extra ASM visitor instances just so as to avoid the thread-local field. I do not see that is should "blow up anyone's face", as @rmannibucau said. The code is simple enough and nothing I am doing there is shady or rocket science. Of course, you can always solve a problem by introducing another layer of complexity, but I have decided against it here and not to cater to someone's personal antipathy for thread-locals. Sorry, but this is my decision for this contribution. We wasted too much time fior discussing this tiny detail already. I would rather have gotten feedback concerning the functionality as such.

@rmannibucau, @eolivelli, would you please also review #88. I have been waiting for a review and for the final merge for too long already. I basically added a new feature for source code shading respecting package exclusions, hence fixing the existing, fragmentary functionality, which was completely broken IMO. Now it is slower, but it does what it should.

kriegaex:
Sorry for the force-push, but I think my change was not worth a separate commit, which would still be visible after merging this (unless the person merging it squashes commits).

src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
rmannibucau: @kriegaex the thread local issue is that it is easily breakable - it is fragile code, wrongly auto-document the code (threadlocal is mainly about multithreaded code and here you have a single thread) and is easily manipulable by code not belonging to this layer (it is not the remapper which owns that logic but the visitor which actually delegates to the writer or writer itself which creates the class). This is why I say it is a quick and dirty fix/workaround and not a future proof solution. Will work on it on monday forking your branch.

kriegaex:
> I wonder if you could try to add a test case in order to prevent regressions in the future

I just added an integration test. Please review and execute.

kriegaex:
BTW, can anyone tell me why running a single IT by

```text
mvn -DskipTests=true -Dinvoker.test=MSHADE-391_noRelocationKeepOriginalClasses verify -P run-its
```

does not work as expected? My test passes, but the setup job fails:

```text
[INFO] --- maven-invoker-plugin:3.2.1:integration-test (integration-test) @ maven-shade-plugin ---
[INFO] Running 1 setup job:
[INFO] Building: setup-parent\pom.xml
[INFO]   The build exited with code 1. See C:\Users\alexa\Documents\java-src\maven-shade-plugin\target\it\setup-parent\build.log for details.
[INFO]           setup-parent\pom.xml ............................. FAILED (2.4 s)
[INFO] Setup done.
[INFO] Building: MSHADE-391_noRelocationKeepOriginalClasses\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           MSHADE-391_noRelocationKeepOriginalClasses\pom.xml SUCCESS (7.3 s)
(...)
[INFO] --- maven-invoker-plugin:3.2.1:verify (integration-test) @ maven-shade-plugin ---
[INFO] -------------------------------------------------
[INFO] Build Summary:
[INFO]   Passed: 1, Failed: 1, Errors: 0, Skipped: 0
[INFO] -------------------------------------------------
[ERROR] The following builds failed:
[ERROR] *  setup-parent\pom.xml
```

If I explicitly add `invoker.test` to the command line

```text
mvn -DskipTests=true -Dinvoker.test=setup-parent,MSHADE-391_noRelocationKeepOriginalClasses verify -P run-its
```

the build passes. That is not so nice, because it requires "sacred knowledge".

src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
kriegaex: Yes, in a way it is premature optimisation. I call it defensive programming. I know that it is single-threaded at the moment, but that could change in the future. The funny thing is, if I would have ignored the thread-unsafety here, nobody would have raised a red flag during the code review. The fact that I did take some precaution, suddenly makes it bad. I disagree with that kind of assessment. But feel free to fork this, if you think you must. I am always happy to learn. I just hink, this should be done in a future PR, not here. I would rather have addressed my question from the TODO comment here, but there the reaction was to do it next time. I guess, different people just have a different sense of priorities, because it is a matter of personal judgement.

I simply disagree with you here about "easily breakable" or "fragile". I am not going to change it.l

rmannibucau: It will likely not change cause multithreading is unlikely to be faster to write a zip until you load all inputs/output in mem which means risking an oom.
Fact encapsulation is not respected and contract is noy consistent with the usage makes me think it is saner to just impl the feature for whta it is rather than doing a quick fix.
Once again, no issue if _you_ do not change it while it is done ;).

kriegaex: > multithreading is unlikely to be faster to write a zip

Good point, I agree. But reading and processing could be done concurrently, writing then queued depending on which thread or pipeline is done with its transformation. I do admit it is a theoretical situation. If after my PR is merged you want to ditch the thread-local in favour of an ordinary boolean (then no longer final in that case, of course) in a subsequent refactoring step, I will not object. The thread-local can always be re-introduced on demand. But at least we have one commit documenting what an (imperfect, but working) solution could look like.

rmannibucau: Well to have tested such options, unless you have tons of relocations (really tons) it will be slower.
Let's wait a few dayd i PR on your pr with my proposal and move forward from that.

Hope it sounds ok.

kriegaex: Actually, after my experience with questions asked on the mailing list, issues opened on Jira and PRs on GitHub, I am kind of afraid that this issue is going to lose momentum and trickle out instead of being merged and closed. Let's not bloat this. We got the code and integration test in place, I would love to see this one merged and then MSHADE-252 next. Still no feedback there, and this is one I really have used for months from my private repository with a temporary version number, because without it Maven Shade is basically useless for me and the AspectJ project, as we also produce shaded and relocated source JARs which should respect excludes, too.

eolivelli (approved):
Thanks for adding the test.

The patch looks great to me

+1

eolivelli:
I am sorry I cannot help with the single run issue.

For some plugjns I modify the pom.xml file in order to select only one single IT.

I have never really worked on this plugin

kriegaex:
FYI, I rebased on master after #88 was merged. Otherwise no changes.

rmannibucau:
Hi @kriegaex , as said last week I had a look to this and have multiple comments to do:

1. I tested the 1 remapper instance per visitor (ClassRemapper) and perfs are not impacted it seems (as expected since the visitor will allocate already way more instance than the class visitor) so I think it can be a good simplicity/perf/design solution actually
2. I tested an alternative to define a custom reduced remapper API to reduce the API surface to what we use (https://github.com/rmannibucau/maven-shade-plugin/commit/50e101feb19f4b82d14346afe613fe6064db99c4), it still relies on "1" (or its variant to set the visitor in the remapper to avoid this allocation but don't think it is worth) but makes the relocating API we use not dependent on ASM which simplifies it a bit at the cost of another interface. Think "1" is simpler but this option has a nice hidden gem: it enables to reimplement ClassRemapper which has limitations in relocations and is inconsistent with our configuration (we dont relocate bytecode but packages only semantically - not sure it is intended but what all the impl do). Don't think we need to go that far for now - once again I'm actually pretty happy with 1 since it is fast and does the work.

So overall, after some more advanced testing on small and "medium/big" (~50M) fatjars with relocations, I think it is good to just store the remapped flag in the remapper and have one instance per class jar entry. It avoids some verbosity we can try to avoid until we do a complete and adapted implementation (which would be way more than this PR tries to do IMHO).

Wdyt?

kriegaex:
@rmannibucau: I showed you mine. If you show me yours, I can say more about it. Is it a PR on top of my PR or a reimplementation of it? I would not be sad to throw away mine, because the sole purpose of it is to fix the problem. If it just remains to be the trigger for you to fix it in a more elegant way, it still served its purpose. So how are we going to do this?

kriegaex:
@rmannibucau, I pushed both your commit and my refactoring of it to this PR, please review again. I hope you are OK with what I changed. The commit comment is quite extensive, if you want to read that first before looking at the diff. I am happy to collaborate with you in this way, that is so much better than a reviewer micro-managing the contributor to do exactly as he says. Taking turns and refining this together is great! I hope we can learn from each other.

Do you think that this one is mergeable now?

rmannibucau (approved):

Tibor17:
@eolivelli 
@rmannibucau 
It looks like you have approved all commits.
Are there any objections to push this?

rmannibucau:
@Tibor17 no, go for it

Tibor17:
@kriegaex Pls squash these commits to one. We will include this PR in a release. Thx

kriegaex:
@Tibor17, I see no reason to squash the commits, because they are doing separate things. I see you use separate commits in one PR too, e.g. lately in Surefire, and I think small commits are a good practice. If ever you need to revert something, chances are that you can revert a single commit instead of a part of a big one. Bisecting bugs is easier, too. Furthermore, in this case there are two separate committers. Squashing everything into one commit would mean to disrespect other people's contribution and copyright. So I am not confident squasing @rmannibucau's commit into mine. The 4 commits so the following:
  1. Adress the core issue
  2. Add an integration test
  3. The other committer refactored something because he did not like my use of thread-local.
  4. I refactored something in his commit because of suboptimal naming.

This code evolution should remain visible and traceable, IMO. If you insist, I can squash 1+2, but I would rather not. Not a big commit history is a problem, big commits are.

rmannibucau:
Not sure it helps but personally I don't care at all if commits are squashed or not, it is up to the code writter to do it (I see it as a doc writer task). Also not having my name in "blame" is fine if you judge it is saner to squash commits.

Tibor17:
@kriegaex 
The reason why it should be squashed is one Jira issue MSHADE-391. The commits are worth for thinking process in the PR but it is useless for the history in master since we do not need to see the reason why this line appeared and it was deleted in another commit within the same issue MSHADE-391.

kriegaex:
@Tibor17, that is not sound reasoning IMO. I completely disagree. Whether it is in a PR or not, later I might need to bisect when there is a problem. Smaller commits mean, it is easier to find bugs. The older a commit gets, the more important it becomes. So commits are more than just part of an iterative development process.

> The reason why it should be squashed is one Jira issue MSHADE-391.

Seriously? That is your argument? Because there is one Jira issue, there must be a single commit? Sorry, I do not mean to disrespect you, but that is complete nonsense.

> The commits are worth for thinking process in the PR but it is useless for the history in master

The opposite is true.

> we do not need to see the reason why this line appeared in one commit and it was deleted in another commit within the same issue MSHADE-391.

Again, I disagree. It helps understand where problems come from and also the opposite, i.e. why some things we would not understand without the history are not problems. Besides, you also want me to squash commits with non-intersecting lines of code.

You know what? If you want to be the dictator micro-managing people, just squash the commits by yourself. I know you think you are always right and you have the power to merge, I don't. But that does not mean I need to always do what you demand. Merge as is or squash by yourself, if it is so important for you, or close the issue. I really don't care, the way you treat contributors like children both here and in Surefire.

Tibor17:
@kriegaex 
Thx for contributing! See the master...

kriegaex:
Thanks for the merge, @Tibor17. Maybe next time you want to retain at least part of the squashed commit comments.

jira-importer:
Resolve #693

PR #128, opened 2022-04-05
[MSHADE-124] keep original basedir after set d-r-p location to build dir

gzm55:
From maven 3.2.5, the pom and basedir are seperated
(https://github.com/apache/maven/commit/c15226f13f6d071d1e926880139d350cc4546b6f),
and add new api `MavenProject::setPomFile()` to only upodate pom location.

In this pr, the minimal maven version is bumped to 3.2.5, then use
`setPomFile` to only update pom file location, leaving `${basedir}`
unchanged.

       pulling in other changes.
       commit message.

### DIFF
--- a/src/it/projects/dep-reduced-pom/pom.xml
+++ b/src/it/projects/dep-reduced-pom/pom.xml
@@ -73,6 +73,28 @@ under the License.
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>3.0.0</version>
+        <executions>
+          <execution>
+            <id>enforce-basedir-is-not-changed</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+	    <phase>package</phase>
+            <configuration>
+              <rules>
+                <requireProperty>
+                  <property>basedir</property>
+                  <regex><![CDATA[.*(?<!/target)$]]></regex>
+                </requireProperty>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/src/it/projects/dep-reduced-pom/pom.xml
+++ b/src/it/projects/dep-reduced-pom/pom.xml
@@ -76,7 +76,7 @@ under the License.
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
-        <version>3.0.0</version>
+        <version>@version.maven-enforcer-plugin@</version>
         <executions>
           <execution>
             <id>enforce-basedir-is-not-changed</id>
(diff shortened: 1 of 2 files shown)


jira-importer:
Resolve #496

jira-importer:
Resolve #496

altaiezior:
@cstamas I wanted to refer my previous comment for this MR, I wanted to change the base location of dependency reduced pom. Can we check this MR and resolves issues if any.

So my use-case here for this fix is to move the dependency-reduced-pom.xml to the target folder, the reason for this that I don't want my search results to include the values from this file.

Target folder is excluded by default in IntelliJ and hence this helps

altaiezior:
@rmannibucau we can just add project.setOriginalModel(model); in this MR after project.setPomFile(dependencyReducedPomLocation); and it should be ready as well and work with maven4

altaiezior:
@cstamas would it be possible to take this up?

cstamas:
@gzm55 care to rebase? Thanks!

altaiezior:
Yes I can but I think I will have to create the MR again, since it is from a fork right? So the history would be lost?

gzm55:
> @gzm55 care to rebase? Thanks!

pr is rebased.

altaiezior:
@cstamas can we check this now?

altaiezior:
@cstamas would you be able to review and merge this change sometime this week?

altaiezior:
Hi @cstamas, just a gentle follow-up on this when you have a moment. Would really appreciate a review whenever you get the chance 🙂

altaiezior:
@cstamas just following up on this PR, if there is anything we can do to get this merged?

cstamas (approved):

slawekjaranowski (approved):

altaiezior:
@cstamas @slawekjaranowski any updates on when this can be released?

slachiewicz:
The fix looks right. `setPomFile` assigns the same `file` field as `setFile` and only skips the `basedir` reassignment, so `getFile()` still returns the reduced POM and install and deploy keep publishing it — the change drops the side effect without changing what ships. It carries `@since 3.2.4`, and this plugin requires 3.6.3, so the floor is clear.

One thing on the new IT guard. The lookbehind tests for `/target`, and a Windows `basedir` ends in `\target`, so the rule passes there whether or not the regression is present:

    .*(?<!/target)$    against  C:\proj\target  ->  matches, rule passes
    .*(?<![/\\]target)$ against C:\proj\target  ->  no match, rule fails

The shared workflow's default `os-matrix` includes `windows-latest`, so a third of the matrix runs the IT without the assertion doing anything. Accepting either separator restores it.

*This comment was created with AI assistance.*

slachiewicz (approved):

slachiewicz:
For whoever picks this up: #827 and #828 propose a different fix for the same problem — saving `project.getFile()` before the dependency-reduced POM is written and restoring it afterwards, rather than this PR's switch from `setFile` to `setPomFile`. They are alternatives, not complements, so one of the three should be chosen before any is merged.

Both of those are currently red, on the Windows cell only, for an unrelated path assumption in their own ITs — the same `/`-only shape as the `basedir` regex noted above.

*This comment was created with AI assistance.*

PR #129, opened 2022-04-06
[MSHADE-323] make shade plugin compatible with the flatten plugin

gzm55:
The runtime injected pom will be set via `MavenProject::setFile` or`MavenProject::setPomFile`, including this plugin and the flatten plugin. So to chain these kind of plugins, when generating the d-r-p pom, this pr first tries to parse a model directly from `MavenProject::getFile` and fallback to a clone of `MavenProject::getOriginalModel` if no pom file set.

Following this checklist to help us incorporate your contribution quickly and easily:

       pulling in other changes.
       commit message.

### DIFF
--- a/src/it/projects/MSHADE-323-compatible-with-flatten-pom/pom.xml
+++ b/src/it/projects/MSHADE-323-compatible-with-flatten-pom/pom.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.its.shade.flatten</groupId>
+  <artifactId>test</artifactId>
+  <version>1.0</version>
+
+  <name>MSHADE-323</name>
+  <description>
+    Test that integration of shade and flatten plugins.
+  </description>
+
+  <properties>
+    <ver>0.1</ver>
+    <maven.jar.forceCreation>true</maven.jar.forceCreation>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven.its.shade.drp</groupId>
+      <artifactId>a</artifactId>
+      <version>${ver}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>flatten-maven-plugin</artifactId>
+        <version>1.2.5</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>flatten</goal>
+            </goals>
+            <phase>prepare-package</phase>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <version>3.2.2</version>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>@project.version@</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
+              <dependencyReducedPomLocation>target/dependency-reduced-pom.xml</dependencyReducedPomLocation>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/src/it/projects/MSHADE-323-compatible-with-flatten-pom/verify.groovy
+++ b/src/it/projects/MSHADE-323-compatible-with-flatten-pom/verify.groovy
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+File pomFile = new File( basedir, "target/dependency-reduced-pom.xml" );
+assert pomFile.isFile()
+
+def ns = new groovy.xml.Namespace("http://maven.apache.org/POM/4.0.0")
+def pom = new XmlParser().parse( pomFile )
+
+assert pom.properties.size() == 0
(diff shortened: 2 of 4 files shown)


src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -1090,7 +1092,10 @@ private void createDependencyReducedPom( Set<String> artifactsToRemove )
             origDeps = transitiveDeps;
         }
 
-        Model model = project.getOriginalModel();
+        final Model model = project.getFile() == null
```
rmannibucau, line 1095: looks like an issue any plugin consuming the pom will get so it should likely be fixed in flatten plugin using `setOriginalModel` to be a global fix and not a local one.

gzm55: hi @rmannibucau , the in Maven, what is 'originalModel` defined to be, a model of the original pom or a model of the current pom file? If defined as the latter one, the global solution of updating originalModel could be done in `MavenProject::setPomFile` or `MavenProject::setFile`.

rmannibucau: My guess is that it was the really original file but since maven is a chain (per module) it is in practise the current pom file - otherwise it can't really be used by any plugins and some do.
So guess the choice is:

1. keep original untouched but means no plugin should use it except help/debug ones (so fix is to always read getFile())
2. ensure flatten plugin updates original model to let downstream plugins consumes the post processing model

Personally I think the option 2 is what makes the most sense for the ecosystem (inter-plugin operability).

gzm55: ok, this pr is the first choice, which prefer a practice single source to get current raw model.

rmannibucau: @gzm55 this pr does not impl 1 since it should never use original model in case 1 but always file to enable to chain plugins.

src/it/projects/MSHADE-323-compatible-with-flatten-pom/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.its.shade.flatten</groupId>
+  <artifactId>test</artifactId>
+  <version>1.0</version>
+
+  <name>MSHADE-323</name>
+  <description>
+    Test that integration of shade and flatten plugins.
```
jglick, line 31:
```suggestion
    Test the integration of Shade and Flatten plugins.
```

src/it/projects/MSHADE-323-compatible-with-flatten-pom/verify.groovy
```diff
@@ -0,0 +1,26 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+File pomFile = new File( basedir, "target/dependency-reduced-pom.xml" );
+assert pomFile.isFile()
+
+def ns = new groovy.xml.Namespace("http://maven.apache.org/POM/4.0.0")
```
jglick, line 23:
```suggestion
def ns = new groovy.xml.Namespace( "http://maven.apache.org/POM/4.0.0" )
```

for consistency

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -1090,7 +1092,10 @@ private void createDependencyReducedPom( Set<String> artifactsToRemove )
...
+                            ? project.getOriginalModel().clone()
+                            : new MavenXpp3Reader().read( new FileInputStream( project.getFile() ) );
```
jglick, line 1097: Possible resource leak?

gzm55: @jglick how to leak the resource?

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
gzm55: it is ok to not fall back to original model when `getFile()` is null, throwing  an exception instead.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rmannibucau: the inputstream will not be closed but the correct way to read a pom is org.apache.maven.model.io.ModelReader anyway

gzm55: thx, let me fix it

rmannibucau: I think we can blame "history" there, xpp3 is kind of old these days ;).

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
rmannibucau: Hmm, I just checked flatten code and it only works if `updatePomFile` is set to true no? So long story short, if you want it to work reliably you must ensure originalModel is set before shade plugin is executed. From a design point of view it is the responsability of flatten plugin to do that since it is the one hacking the user pom but an intermediary plugin setting it from a file when flatten plugin dumped its value would work too but I don't think there is any bug in shade there.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
gzm55: ok, now we use ModelReader to parse the pom

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
gzm55: it seems better to have a convention or rule that the original model should be equal to the pom file. the shade plugin also breaking this convention, it reverts the parent.relativePath after write the model to the file.

rmannibucau: shade plugins reset it cause it just needs to change it while dumping the pom (that said good point it breaks thread safety in concurrent builds).
agree shade plugin breaks this rules and happy to get it fixed, it is just not that important cause shade plugin is generally in the end of the chain (so dowstream consumers list is empty) and it also handles itself the pom rewriting but you are right.
long story short: I don't think we should re-read from the filesystem any pom model anytime in maven build otherwise the design itself is broken so I'm in favor of trying to not use the reader more than once per file.

hope it makes sense.

jira-importer:
Resolve #323

jira-importer:
Resolve #323

anuragagarwal561994:
@olamy @slawekjaranowski  can we check if this can be deployed in the next version?

altaiezior:
@cstamas would be help get this fix in the next release? This is an older but simpler fix.

So my use-case here for this fix is to move the dependency-reduced-pom.xml to the target folder, the reason for this that I don't want my search results to include the values from this file.

Target folder is excluded by default in IntelliJ and hence this helps

rmannibucau:
@altaiezior not sure about AI but the model can be injected and we have a writer in the codebase so shouldn't be that hard.
My 2 cts would be that since maven 4 has flattening feature built-in it can be worth testing it and working with maven 4 more than maven 3 base now - but API changed, this is why I think we can  just play the status quo to only update maven 4 plugin code base once setup.

altaiezior:
> TBH, I don't quite the problem here. PR is about "chaining flatten plugin and this plugin", and you mention IDE and search... Can someone enlighten me more?

Yes you are write, actually my comment was meant for https://github.com/apache/maven-shade-plugin/pull/128, I did mistakenly added here.

altaiezior:
Actually I re-validated that the latest version of flatten-maven-plugin and maven-shade-plugin are working together.

I was using flatten-maven-plugin to use the ci-friendly version and earlier I was facing the issue where the dependency--reduced-pom.xml was still having the ${revision} tag so I had to run the flatten plugin after the maven-shade-plugin.

But I did a couple of changes with my original configuration of the pom.xml and now everything seems to be working fine. So I am not sure if the original issue was fixed and is this required anymore.

altaiezior:
> @altaiezior not sure about AI but the model can be injected and we have a writer in the codebase so shouldn't be that hard. My 2 cts would be that since maven 4 has flattening feature built-in it can be worth testing it and working with maven 4 more than maven 3 base now - but API changed, this is why I think we can just play the status quo to only update maven 4 plugin code base once setup.

Yes I am also testing with maven-4 and there has been no issues with maven-4 as we don't use flatten plugin with the same.

jglick:
Regarding testing with Maven 4, beware https://github.com/apache/maven/pull/11370

altaiezior:
I re-checked it with maven 4, actually the dependency reduced pom is being considered as the build pom and not the consumer pom so the final pom being installed is not dependency reduced and hence it is not working with it.

I was going through the discussion here https://github.com/apache/maven/pull/1437 which helps me understand why changing the model is important.

In short, the consumer poms are generated from the model and not the file. So if we are changing and creating a new pom file by any plugin, then one needs to also change the model or else the consumer pom will not be updated.

altaiezior:
@rmannibucau I think this issue can be closed as flatten-maven-plugin now implements setting originalModel which can then be consumed by maven-shade-plugin

https://github.com/mojohaus/flatten-maven-plugin/commit/f85967aa8f965579f6796661de555f3a777cfc4e

when updatePomFile property is set

PR #149, opened 2022-08-31
[MSHADE-419] createDependencyReducedPom should not replace original pom when shadedArtifactAttached

XN137:
### Description:

The problem was introduced by [MSHADE-321](https://github.com/apache/maven-shade-plugin/commit/6ea8543b8b380bd4ebc0e876970957426bb46808) in 3.3.0 because before,
the `createDependencyReducedPom` method was only ever called when
`shadedArtifactAttached` is false.

The `createDependencyReducedPom` method not only writes the dependency
reduced pom file but also modifies the current project to use it as
replacement for its original pom.
This has an effect on all maven build logic being executed afterwards.

Thus, in multi module builds, when a module attaches a shaded jar, its
non-shaded artifact no longer contains transitive dependencies when used
from other modules, even though the pom in its jar still declares them.

### Checklist

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MSHADE-419) filed 
       pulling in other changes.
       commit message.

### DIFF
--- a/src/it/projects/MSHADE-419_createDependencyReducedPom_multimodule/api/pom.xml
+++ b/src/it/projects/MSHADE-419_createDependencyReducedPom_multimodule/api/pom.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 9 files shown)


XN137:
@khmarbaise sorry for the ping, but whats the best way to find someone to review this? thank you

cstamas (approved):

gnodet:
The problem seems to have been fixed in 3.4.1 according to #162 which seems to pass on master.

frant-hartm:
@gnodet the test here and in #162 is incomplete - it should also check what pom was installed into the local repository.

jira-importer:
Resolve #711

jira-importer:
Resolve #711

PR #162, opened 2022-10-22
[MSHADE-419] Add an IT for MSHADE-419

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MSHADE-419

Shade plugin causes pom to be created without compile dependencies

IT extracted from #149 

===============

The problem was introduced by MSHADE-321 in 3.3.0 because before, the `createDependencyReducedPom` method was only ever called when `shadedArtifactAttached` is false.

The `createDependencyReducedPom` method not only writes the dependency reduced pom file but also modifies the current project to use it as replacement for its original pom.
This has an effect on all maven build logic being executed afterwards.

Thus, in multi module builds, when a module attaches a shaded jar, its non-shaded artifact no longer contains transitive dependencies when used from other modules, even though the pom in its jar still declares them.

# Conflicts:
#	src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java

       pulling in other changes.
       commit message.

### DIFF
--- a/src/it/projects/MSHADE-419_createDependencyReducedPom_multimodule/verify.groovy
+++ b/src/it/projects/MSHADE-419_createDependencyReducedPom_multimodule/verify.groovy
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.jar.*
+import org.codehaus.plexus.util.*
+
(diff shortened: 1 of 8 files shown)


src/it/projects/MSHADE-419_createDependencyReducedPom_multimodule/verify.groovy
```diff
@@ -0,0 +1,61 @@
...
+}
+
+String reducedPomFileContent = FileUtils.fileRead( new File( basedir, "impl/dependency-reduced-pom.xml" ), "UTF-8" )
+if ( reducedPomFileContent.contains( apiDependencyString ) )
+{
+    throw new IllegalStateException( "The dependency-reduced-pom.xml should not contain the api dependency:\n" + reducedPomFileContent )
+}
+
+JarFile jarFile = null
+try
```
elharo, line 45: I think you can use try with resources here

src/it/projects/MSHADE-419_createDependencyReducedPom_multimodule/verify.groovy
```diff
@@ -0,0 +1,61 @@
...
+// note that the following checks were passing even before any fix for MSHADE-419,
+// since the generated poms and jars are generated correctly.
+// only compilation of the impl-user module is failing because it uses the normal jar but
+// without its transitive dependencies
+
+String jarPrefix = "impl/target/mshade-419-impl-1.0"
+String jarPomPath = "META-INF/maven/org.apache.maven.its.shade.cdrp-mm/mshade-419-impl/pom.xml"
+String apiDependencyString = "<artifactId>mshade-419-api</artifactId>"
+
+String mainPomFileContent = FileUtils.fileRead( new File( basedir, "impl/pom.xml" ), "UTF-8" )
```
elharo, line 32: can we use the Apache Commons IO version instead?

jira-importer:
Resolve #711

jira-importer:
Resolve #711

PR #171, opened 2022-12-30
Update overlapping resources warning

pzygielo:


rmannibucau (approved):
(thinking out loud: maybe we can just use "file" instead of "class or resource")

pzygielo:
Thanks for checking. Probably just "file" would be even better indeed!

pzygielo:
May I ask for the merge, please?

slachiewicz (approved):

PR #175, opened 2023-02-22
[MSHADE-438] Update to Maven 3.2.5

cstamas:
Up the baseline as for the rest of ASF plugins.

Changes:
* drop MAT
* drop uses of archaic constructs

Also update IT for MSHADE-340 as it was not passing with maven-jar-plugin 3+ version (maven 3.8.x used maven-jar-plugin 2.4), as the IT was replacing the main JAR intentionally, or unintentionally.

---

https://issues.apache.org/jira/browse/MSHADE-438

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -19,15 +19,14 @@ package org.apache.maven.plugins.shade.mojo;
  * under the License.
  */
 
+import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.Artifact;
-import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Exclusion;
 import org.apache.maven.model.Model;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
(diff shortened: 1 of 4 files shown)


michael-o (approved):
Very nice. Can't believe we still haven't upgraded to 3.2.5.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -846,29 +833,29 @@ private void replaceFile( File oldFile, File newFile )
...
 
     private File resolveArtifactForClassifier( Artifact artifact, String classifier )
     {
-        DefaultArtifactCoordinate coordinate = new DefaultArtifactCoordinate();
-        coordinate.setGroupId( artifact.getGroupId() );
-        coordinate.setArtifactId( artifact.getArtifactId() );
-        coordinate.setVersion( artifact.getVersion() );
-        coordinate.setExtension( "jar" );
-        coordinate.setClassifier( classifier );
+        ArtifactType artifactType = session.getRepositorySession().getArtifactTypeRegistry().get( artifact.getType() );
```
cstamas, line 845: This is NPE prone: i forgot that resolver ArtifactTypeRegistry is DIFFERENT that maven's ArtifactHandlerManager (never returns null).

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -846,29 +833,29 @@ private void replaceFile( File oldFile, File newFile )
...
+        org.eclipse.aether.artifact.DefaultArtifact coordinate = new DefaultArtifact( artifact.getGroupId(),
+                artifact.getArtifactId(), classifier, artifactType.getExtension(), artifact.getVersion(),
```
cstamas, line 847: actually this `artifactType.getExtension()` is NPE prone, as artifactType can be null

cstamas:
> Very nice. Can't believe we still haven't upgraded to 3.2.5.

btw, this one catches nicely all these :wink: https://issues.apache.org/jira/browse/MNG-7706

slawekjaranowski (approved):

jira-importer:
Resolve #665

PR #188, opened 2023-06-13
[MSHADE-447] Do not use forked guice, use sisu from parent

cstamas:
Use proper Guice, same as Maven. Drop sisu plugin, is defined in parent.

---

https://issues.apache.org/jira/browse/MSHADE-447


jira-importer:
Resolve #707

PR #196, opened 2023-09-19
Adjust to test also with Java 21

slachiewicz:


pom.xml
```diff
@@ -365,5 +366,44 @@
...
+          <plugin>
+            <groupId>com.diffplug.spotless</groupId>
+            <artifactId>spotless-maven-plugin</artifactId>
+            <configuration>
+              <java>
+                <palantirJavaFormat>
+                  <version>2.35.0</version>
+                </palantirJavaFormat>
+              </java>
+            </configuration>
```
slawekjaranowski, line 403: Should be done also in parent 😄

slawekjaranowski:
Should be linked with: https://issues.apache.org/jira/browse/MSHADE-45

pom.xml
slachiewicz: I'm waiting for the sportless plugin will be released with the complete fix. Now just experimenting locally only.

slawekjaranowski (approved):

PR #210, opened 2024-01-20
[MSHADE-467] Fix concurrency problem with dependency-reduced POM

kriegaex:
Fixes https://issues.apache.org/jira/browse/MSHADE-467.

You review the fix. I converted the reproducer project into the integration test from the first commit. The second commit makes the previously failing test pass.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -35,6 +35,8 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.Artifact;
@@ -1048,6 +1050,8 @@ public class ShadeMojo extends AbstractMojo {
         rewriteDependencyReducedPomIfWeHaveReduction(dependencies, modified, transitiveDeps, model);
     }
 
+    private static final Lock LOCK = new ReentrantLock();
+
     private void rewriteDependencyReducedPomIfWeHaveReduction(
             List<Dependency> dependencies, boolean modified, List<Dependency> transitiveDeps, Model model)
             throws IOException, ProjectBuildingException, DependencyGraphBuilderException {
@@ -1112,15 +1116,21 @@ public class ShadeMojo extends AbstractMojo {
                     w.close();
                 }
 
-                ProjectBuildingRequest projectBuildingRequest =
-                        new DefaultProjectBuildingRequest(session.getProjectBuildingRequest());
-                projectBuildingRequest.setLocalRepository(session.getLocalRepository());
-                projectBuildingRequest.setRemoteRepositories(project.getRemoteArtifactRepositories());
+                // Lock critical section to fix MSHADE-467
+                try {
+                    LOCK.lock();
+                    ProjectBuildingRequest projectBuildingRequest =
+                            new DefaultProjectBuildingRequest(session.getProjectBuildingRequest());
+                    projectBuildingRequest.setLocalRepository(session.getLocalRepository());
+                    projectBuildingRequest.setRemoteRepositories(project.getRemoteArtifactRepositories());
 
-                ProjectBuildingResult result = projectBuilder.build(f, projectBuildingRequest);
+                    ProjectBuildingResult result = projectBuilder.build(f, projectBuildingRequest);
 
-                getLog().debug("updateExcludesInDeps()");
-                modified = updateExcludesInDeps(result.getProject(), dependencies, transitiveDeps);
+                    getLog().debug("updateExcludesInDeps()");
+                    modified = updateExcludesInDeps(result.getProject(), dependencies, transitiveDeps);
+                } finally {
+                    LOCK.unlock();
+                }
             }
 
             project.setFile(dependencyReducedPomLocation);
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -35,8 +35,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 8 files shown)


kriegaex:
IT is in place now after a force-push. Please review.

kriegaex:
Hm, according to https://github.com/kriegaex/maven-shade-plugin/actions/runs/7592589521, the fix does not work universally on all combinations of Maven and JDK versions. I am open for input.

kriegaex:
OK, [this build](https://github.com/kriegaex/maven-shade-plugin/actions/runs/7593219673) has passed now. For Maven 3.9.6, synchronisation on a single session instance was enough, but for Maven 3.6.3 it was a bit too naive and optimistic. Complex objects like project, session etc. contain lots of nested stuff, and not everything can or even should be deep-cloned as was done in MSHADE-413 in another part of the mojo. That can work, but leads to potential brittle solutions, if the plugin needs to run on multiple Maven versions.

So, protection of the critical section with a `ReentrantLock` it is, then.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -1112,15 +1116,21 @@ private void rewriteDependencyReducedPomIfWeHaveReduction(
                     w.close();
                 }
 
-                ProjectBuildingRequest projectBuildingRequest =
-                        new DefaultProjectBuildingRequest(session.getProjectBuildingRequest());
-                projectBuildingRequest.setLocalRepository(session.getLocalRepository());
-                projectBuildingRequest.setRemoteRepositories(project.getRemoteArtifactRepositories());
+                // Lock critical section to fix MSHADE-467
+                try {
```
rmannibucau, line 1120: A global lock looks wrong, shouldnt we lock on the project (we can store a concurrenthashmap of project/lock in the session to avoid synchronized for the future if we enable virtual threads a day)

kriegaex: I tried, it is not enough. Feel free to suggest something better, but after you verified that it works, please.

kriegaex: BTW, feel free to commit on top of the PR, we can always squash some commits later, as long as we keep the IT separate from the fix.

image

rmannibucau:
```
diff --git a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
index 3a7eaaa..a8475d9 100644
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -35,8 +35,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.Artifact;
@@ -1050,8 +1048,6 @@ public class ShadeMojo extends AbstractMojo {
         rewriteDependencyReducedPomIfWeHaveReduction(dependencies, modified, transitiveDeps, model);
     }
 
-    private static final Lock LOCK = new ReentrantLock();
-
     private void rewriteDependencyReducedPomIfWeHaveReduction(
             List<Dependency> dependencies, boolean modified, List<Dependency> transitiveDeps, Model model)
             throws IOException, ProjectBuildingException, DependencyGraphBuilderException {
@@ -1116,9 +1112,7 @@ public class ShadeMojo extends AbstractMojo {
                     w.close();
                 }
 
-                // Lock critical section to fix MSHADE-467
-                try {
-                    LOCK.lock();
+                synchronized (session.getProjectBuildingRequest()) { // Lock critical section to fix MSHADE-467
                     ProjectBuildingRequest projectBuildingRequest =
                             new DefaultProjectBuildingRequest(session.getProjectBuildingRequest());
                     projectBuildingRequest.setLocalRepository(session.getLocalRepository());
@@ -1128,8 +1122,6 @@ public class ShadeMojo extends AbstractMojo {
 
                     getLog().debug("updateExcludesInDeps()");
                     modified = updateExcludesInDeps(result.getProject(), dependencies, transitiveDeps);
-                } finally {
-                    LOCK.unlock();
                 }
             }
 

```

can you give a try to that and check it is sufficient for you please? should fix your issue without putting a global bottleneck in the mojo (previous comment is still accurate, no strong opinion on lock vs synchronized as of today, just want a bottleneck per project to avoid a bottleneck in complex projects).

kriegaex: LGTM, thanks for the great input. I made sure your authorship was respected in the latest commit. The CI build on my fork is passing. Before merging, would you like me to squash our two solutions into one commit? I wwould prefer to keep both commits for reference and to see the code evolution. Should the necessity for a reentrant lock arise in the future, at least we have an example. We also have one already for what is generally preferable, like in this case. As I cannot merge myself, I leave it up to you to decide.

rmannibucau: Looks ok to me like that - if another commiter is ok then let's go ;)

fmarot:
Big big thank you for your work @kriegaex 👍

kriegaex:
I do not think it is pure chance that it happens in the new IT. Maybe, in some other part of Maven, at least in some versions, there are other concurrency issues we are hitting here by chance, just because we are testing this setup. If you run the build a few times and it does not re-occur, maybe the test is "just" flaky because of Maven or some situation on the GitHub server. If it happens more often, maybe we can try my initial solution with the reentrant lock and see if that one produces more stable results.

Looking at the stack trace below the [error message in the Maven 3.6.3 JDK 8 Windows build](https://github.com/apache/maven-shade-plugin/actions/runs/7596727600/job/20718211260#step:8:8504), it seems as if the Maven `MultiThreadedBuilder` is involved. Maybe its resource management is suboptimal on a platform like Windows where open files are locked.

kriegaex:
@stl543, @fmarot, you are welcome. But looking at the commits, what effectively ended up in the merge were basically your reproducer and a variation of my fix provided by @rmannibucau. I.e., almost all the credit stays in France. I just debugged and connected the dots. Allez les Bleus! 🇫🇷

jira-importer:
Resolve #721

jira-importer:
Resolve #721

PR #213, opened 2024-02-01
[MSHADE-468] add system requirements history

hboutemy:
https://issues.apache.org/jira/browse/MSHADE-468

notice the use of plugin versions ranges "from x to y" to avoid showing the whole plugin releases history while explaining the range of releases sharing the same prerequisites

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -300,6 +300,54 @@
             </execution>
           </executions>
         </plugin>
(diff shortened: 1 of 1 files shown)


slawekjaranowski (approved):

jira-importer:
Resolve #731

PR #219, opened 2024-04-05
Add `defaultValue` for all boolean Mojo parameters

Marcono1234:
Currently the Mojo documentation does not mention the default for these parameters, which will be confusing for the user. Have for example a look at the current https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html documentation; you will see that for all the changed parameters no default value is currently documented. So it is not clear whether these parameters apply by default.

Please let me know if I forget to consider something and this change has any undesired side effects.

----

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -197,7 +197,7 @@ public class ShadeMojo extends AbstractMojo {
      * <p/>
      * If you like to change the name of the native artifact, you may use the &lt;build>&lt;finalName> setting. If this
      * is set to something different than &lt;build>&lt;finalName>, no file replacement will be performed, even if
-     * shadedArtifactAttached is being used.
+     * {@link #shadedArtifactAttached} is false.
      */
     @Parameter
     private String finalName;
@@ -217,10 +217,10 @@ public class ShadeMojo extends AbstractMojo {
     private String shadedGroupFilter;
 
     /**
-     * Defines whether the shaded artifact should be attached as classifier to the original artifact. If false, the
-     * shaded jar will be the main artifact of the project
+     * Defines whether the shaded artifact should be attached as classifier to the original artifact. If false (the
+     * default), the shaded jar will be the main artifact of the project.
      */
-    @Parameter
+    @Parameter(defaultValue = "false")
     private boolean shadedArtifactAttached;
 
     /**
@@ -265,14 +265,14 @@ public class ShadeMojo extends AbstractMojo {
     /**
      * When true, dependencies are kept in the pom but with scope 'provided'; when false, the dependency is removed.
      */
-    @Parameter
+    @Parameter(defaultValue = "false")
     private boolean keepDependenciesWithProvidedScope;
 
     /**
      * When true, transitive deps of removed dependencies are promoted to direct dependencies. This should allow the
      * drop in replacement of the removed deps with the new shaded jar and everything should still work.
      */
-    @Parameter
+    @Parameter(defaultValue = "false")
     private boolean promoteTransitiveDependencies;
 
     /**
@@ -284,13 +284,13 @@ public class ShadeMojo extends AbstractMojo {
     /**
      * When true, it will attempt to create a sources jar as well
      */
-    @Parameter
+    @Parameter(defaultValue = "false")
     private boolean createSourcesJar;
 
     /**
      * When true, it will attempt to create a test sources jar.
      */
-    @Parameter
+    @Parameter(defaultValue = "false")
     private boolean createTestSourcesJar;
 
     /**
@@ -321,7 +321,7 @@ public class ShadeMojo extends AbstractMojo {
      *
      * @since 1.4
      */
-    @Parameter
+    @Parameter(defaultValue = "false")
     private boolean minimizeJar;
 
     /**


src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -197,7 +197,7 @@ public class ShadeMojo extends AbstractMojo {
      * <p/>
      * If you like to change the name of the native artifact, you may use the &lt;build>&lt;finalName> setting. If this
      * is set to something different than &lt;build>&lt;finalName>, no file replacement will be performed, even if
-     * shadedArtifactAttached is being used.
+     * {@link #shadedArtifactAttached} is false.
```
Marcono1234, line 200: Is this change correct? If I understood it correctly, if `<build><finalName>` and the Mojo `<finalName>` differ then the main project artifact will not be replaced, even if `shadedArtifactAttached=false`.

hboutemy: to me, "shadedArtifactAttached is being used" means setting is t to `true`

hboutemy: this part of the PR should IMHO just be dropped

Marcono1234: Ok, I have reverted that change.

slachiewicz (approved):

rmannibucau:
Open question: fix pluginxml generation instead of fixing all mojo since it is valid to use the default value for primitives?

Marcono1234:
That might indeed be a better solution. Can there be cases though were a parameter has a different default value, which is not set with `@Parameter(defaultValue = "...")`?

For example, if the constructor of a Mojo assigns a value to a field. Does that act as default value then?

rmannibucau:
@Marcono1234 it is possible but will not be in the documentation (requires textual description) but guess the good practise stays to write the less so if default is automatic doc should be aware of it somehow, either ensuring we inject the default for primitives in plugin.xml or let the doc handle it IMHO.

Marcono1234:
> doc should be aware of it somehow, either ensuring we inject the default for primitives in plugin.xml or let the doc handle it IMHO

But how do you propose this should be done without being error-prone? Or do you propose to not only improve how the documentation is generated, but also how Maven determines the `@Parameter` default parameter?

Another case which comes to my mind is:
```java
@Parameter
boolean param = true;
```

Will the documentation plugin be aware that the default value is `true`? It would either need to parse the source code or analyze the byte code.

Having the documentation plugin just document `false` for all `@Parameter boolean` without `defaultValue = "..."` would be incorrect it seems.

----

So yes, maybe this issue could be solved generally for all plugins by doing a change in Maven Plugin Plugin (?). But to me it seems the solution is not simple (if possible at all).

So would you mind accepting this pull request here the way it is for now?

rmannibucau:
The 2 simple options:

* Enhance plugin.xml to set defaults for primitives
* Enhance doc generation to handle primitives

(I prefer 1 since it will work for all consumers but 2 works for "us")

The inline init case is not supposed to work and it is also hard since init can be conditional (env for ex) so let's ignore coded init and just stick to implicit defaults.

Marcono1234:
> The inline init case is not supposed to work

But does it work currently, and are plugins relying on this? Because if Maven Plugin Plugin is changed to pick a different default value in the future in those cases, or if it starts documenting a default value which is not actually the default value, then I doubt that those changes will be made.

Either way, I think this is getting a bit off-topic. Would you mind merging the changes here?
Even if Maven Plugin Plugin is changed, that will happen at some point in the future, and until then the Shade plugin documentation would still be incomplete.

If you want the Maven Plugin Plugin to change, could you or someone else please drive those changes, because I am neither familiar enough with Maven nor Maven plugin development or the internals, and I personally also doubt that such a change would be worth it given the compatibility risks.

rmannibucau:
Some plugings use it but it since it is not static we cant handle it automatically.

I'd prefer to fix the doc and not pollute shade plugin with unexpected code if possible but just my 2cts.

Marcono1234:
@rmannibucau, I had a look at this again, but I still don't see how Maven Plugin Plugin or similar could be enhanced to infer the default value. And suddenly determining a different default value would break existing plugins. See my earlier comment above: https://github.com/apache/maven-shade-plugin/pull/219#issuecomment-2143611680

This pattern of assigning the default value using the field initializer is not that uncommon. Examples:
- [OpenAPITools/openapi-generator](https://github.com/OpenAPITools/openapi-generator/blob/53f1094765408a1b6f12fb3f8570f5e61e876600/modules/openapi-generator-maven-plugin/src/main/java/org/openapitools/codegen/plugin/CodeGenMojo.java#L410-L411)
- [GoogleContainerTools/jib](https://github.com/GoogleContainerTools/jib/blob/ff15988e5eb36c686e7db47abb31bdc15c88f44a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java#L376-L377)
- [quarkusio/quarkus-platform](https://github.com/quarkusio/quarkus-platform/blob/562e7e02c1fb67ab4b254cd91bb98a36fde4201a/generated-platform-project/quarkus-maven-plugin/src/main/java/io/quarkus/maven/DevMojo.java#L342-L343)
- [apache/camel-spring-boot](https://github.com/apache/camel-spring-boot/blob/5c807b27a67c6a6f6b1693806a70791449a828ec/tooling/camel-spring-boot-generator-maven-plugin/src/main/java/org/apache/camel/springboot/maven/BomDependenciesGeneratorMojo.java#L147-L148)
- [flyway/flyway](https://github.com/flyway/flyway/blob/f608dcade2a56a4717aedb78aa37f3acafd7352f/flyway-maven-plugin/src/main/java/org/flywaydb/maven/AbstractFlywayMojo.java#L645-L646)

Marcono1234:
@slachiewicz, would you mind merging these changes?

These changes hopefully help users _now_ (respectively the next Shade Plugin release), whereas some general Maven Plugin Plugin changes (in case they are ever performed, see my concerns above) might only happen after months or years.

PR #226, opened 2024-05-27
[MSHADE-478] Extra JARs feature

cstamas:
Ability to add "extra JARs" and still enjoy the full benefits of relocation, resource transformation, etc.

---

https://issues.apache.org/jira/browse/MSHADE-478

---

PS: This change was done against https://github.com/apache/maven-shade-plugin/pull/225 as part of 3.6.0 changes.


cstamas:
This is not enough, filters are not applied. Reason is that artifact keys are not added to very convoluted structure of filters...

cstamas:
Superseded by https://github.com/apache/maven-shade-plugin/pull/228

jira-importer:
Resolve #675

PR #228, opened 2024-05-28
[MSHADE-478] Extra JARs feature

cstamas:
Ability to add "extra JARs and Artifacts" and still enjoy the full benefits of relocation, resource transformation, etc.

---

https://issues.apache.org/jira/browse/MSHADE-478

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -396,9 +396,8 @@ public class ShadeMojo extends AbstractMojo {
      * Extra JAR files to infuse into shaded result. Accepts list of files that must exists. If any of specified
      * files does not exist (or is not a file), Mojo will fail.
      * <p>
-     * One can add here "extra JARs", to be worked into the resulting shaded JAR. The listed JAR files must exist.
-     * Extra JARs will be processed in same way as main JAR (if any): applied relocation, resource transformers
-     * but not filtering.
+     * Extra JARs will be processed in same way as main JAR (if any) is: applied relocation, resource transformers
+     * but <em>not filtering</em>.
      * <p>
      * Note: this feature should be used lightly, is not meant as ability to replace dependency hull! It is more
      * just a feature to be able to slightly "differentiate" shaded JAR from main only.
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -393,6 +392,36 @@ public class ShadeMojo extends AbstractMojo {
...
+     * Extra JARs will be processed in same way as main JAR (if any): applied relocation, resource transformers
+     * but not filtering.
+     * <p>
+     * Note: this feature should be used lightly, is not meant as ability to replace dependency hull! It is more
+     * just a feature to be able to slightly "differentiate" shaded JAR from main only.
+     *
+     * @since 3.6.0
+     */
+    @Parameter
+    private List<File> extraJars;
```
slawekjaranowski, line 408: we should add what is the item ... I guess that is a path to jar file

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -393,6 +392,36 @@ public class ShadeMojo extends AbstractMojo {
...
+     * (not transitively), and will be processed in same way as dependency JARs (if any): regarding relocation,
+     * resource transformers and filtering.
+     * <p>
+     * Note: this feature should be used lightly, is not meant as ability to replace dependency hull! It is more
+     * just a feature to be able to slightly "differentiate" shaded JAR from main only.
+     *
+     * @since 3.6.0
+     */
+    @Parameter
+    private List<String> extraArtifacts;
```
slawekjaranowski, line 423: We should describe format of item ... `groupId:artifactId:...` 
which items are required?

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
cstamas: Yes, these are files.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
cstamas: The ["standard" format](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/artifact/DefaultArtifact.java#L51-L63), will add to comment.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
cstamas: pushed update

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -444,6 +474,17 @@ public void execute() throws MojoExecutionException {
 
             artifacts.add(project.getArtifact().getFile());
 
+            if (extraJars != null && !extraJars.isEmpty()) {
+                for (File extraJar : extraJars) {
+                    if (!Files.isRegularFile(extraJar.toPath())) {
+                        createErrorOutput();
```
slawekjaranowski, line 480: Please check if this error message are correct here 😄

slawekjaranowski (approved):

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
slawekjaranowski:
```java
private void createErrorOutput() {
        getLog().error("The project main artifact does not exist. This could have the following");
        getLog().error("reasons:");
....
```

slawekjaranowski (approved):

jira-importer:
Resolve #675

jira-importer:
Resolve #675

PR #229, opened 2024-05-28
[MSHADE-428] Prevent null value in array of transformers

cstamas:
As Plexus will silently add it as explain it in issue.

---

https://issues.apache.org/jira/browse/MSHADE-428


jira-importer:
Resolve #727

PR #744, opened 2025-08-07
Bump ASM 9.8 to support JDK 25 bytecode

pan3793:
The current ASM 9.7 does not support JDK 24 and 25 bytecode.

```
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.6.0:shade (default) on project kyuubi-spark-sql-engine_2.12: Error creating shaded jar: Problem shading JAR /Users/chengpan/.m2/repository/net/bytebuddy/byte-buddy/1.17.6/byte-buddy-1.17.6.jar entry META-INF/versions/24/net/bytebuddy/jar/asmjdkbridge/JdkClassWriter$1.class: java.lang.IllegalArgumentException: Unsupported class file major version 68 -> [Help 1]
...
```
---

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
     <mavenVersion>3.9.7</mavenVersion>
     <javaVersion>8</javaVersion>
     <currentVersion>${project.version}</currentVersion>
-    <asmVersion>9.7</asmVersion>
+    <asmVersion>9.8</asmVersion>
     <slf4j.version>1.7.36</slf4j.version>
     <project.build.outputTimestamp>2024-05-28T15:16:00Z</project.build.outputTimestamp>
   </properties>


InsertPikachuFace:
@pan3793 @olamy Just to be clear, is version 3.6.1 compatible with JDK 25 or just 24? Because I get this error for a JDK 25 project-build:
> Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.6.1:shade (default) on project kfz-fahrzeugsuche-service: Error creating shaded jar: Unsupported class file major version 69

If not, how would one "override the asm dependency in your pom" to work with JDK 25?

pan3793:
@InsertPikachuFace I'm not much familiar with this plugin, below answers are based on my understanding.

> is version 3.6.1 compatible with JDK 25 or just 24?

It should support JDK 25.

> how would one "override the asm dependency in your pom" to work with JDK 25?

Apache Kyuubi resolved the issue via https://github.com/apache/kyuubi/pull/7164

Issue #573, opened 2015-08-02
[MSHADE-202] When promoteTransitiveDependencies=true, some <exclusions> are stripped from the dependency-reduced-pom


jira-importer:
**[Andrey Stepaonv](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=drutech)** opened **[MSHADE-202](https://issues.apache.org/jira/browse/MSHADE-202?redirect=false)** and commented

I have 2 modules. First module uses this dependencies:

```xml
<dependency>
    <groupId>ch.imvs</groupId>
    <artifactId>sdes4j</artifactId>
    <version>1.1.3</version>
    <exclusions>
        <exclusion>
            <groupId>commons-codec</groupId>
            <artifactId>commons-codec</artifactId>
        </exclusion>
    </exclusions>
</dependency>
<dependency>
    <groupId>commons-codec</groupId>
    <artifactId>commons-codec</artifactId>
    <version>1.10</version>
</dependency>
```

and second module uses first module and maven-shade plugin with execution:

```xml
<execution>
<id>shade-for-obfuscate</id>
<phase>package</phase>
<goals>
    <goal>shade</goal>
</goals>
<configuration>
    <artifactSet>
        <includes>
            <include>${project.groupId}:firstModule</include>
        </includes>
    </artifactSet>
    <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
    <dependencyReducedPomLocation>${basedir}/target/dependency-reduced-pom.xml</dependencyReducedPomLocation>
</configuration>
</execution>
```

As a result i have reduced pom with dependencies:

```xml
<dependency>
  <groupId>ch.imvs</groupId>
  <artifactId>sdes4j</artifactId>
  <version>1.1.3</version>
  <scope>compile</scope>
</dependency>
<dependency>
  <groupId>commons-codec</groupId>
  <artifactId>commons-codec</artifactId>
  <version>1.10</version>
  <scope>compile</scope>
</dependency>
```

But i had to get dependency with \<exclusions>

---

**Affects:** 2.4.1

**Issue Links:**
- [MSHARED-852](https://issues.apache.org/jira/browse/MSHARED-852) Add Exclusions to DependencyNode
 (_**"depends upon"**_)

1 votes, 3 watchers

PR #770, opened 2025-11-09
[MSHADE-573] Fix missing exclusions - direct and transitive 

marwin1991:

       pulling in other changes. =https://github.com/apache/maven-shade-plugin/issues/573
       commit message.

 ## Pull Request Description

Preserve original and transitive exclusions in dependency-reduced-pom to improve POM correctness

### Summary
This PR fixes a long-standing issue where exclusions declared in the original `pom.xml` — including wildcard and transitive exclusions — were not propagated to the generated `dependency-reduced-pom.xml`. As a result, the reduced POM could reintroduce dependencies that the user had explicitly excluded.

The change introduces a dedicated exclusion resolution step and updates the reduced-POM generation so that:
- Direct exclusions defined on dependencies in the original POM are preserved.
- Transitive exclusions (required to keep the effective dependency graph consistent) are carried over.

### Motivation and context
Example of the incorrect behavior (existing upstream case):
- Source POM defines a wildcard exclusion on a dependency:
  ```xml
  <dependency>
      <groupId>com.itextpdf</groupId>
      <artifactId>itextpdf</artifactId>
      <version>5.5.13.3</version>
      <exclusions>
          <exclusion>
              <groupId>*</groupId>
              <artifactId>*</artifactId>
          </exclusion>
      </exclusions>
  </dependency>
  ```
- But the resulting `dependency-reduced-pom.xml` previously dropped the exclusions entirely:
  ```xml
  <dependency>
      <groupId>com.itextpdf</groupId>
      <artifactId>itextpdf</artifactId>
      <version>5.5.13.3</version>
  </dependency>
  ```

Reference: https://github.com/apache/maven-shade-plugin/blob/073e7ca970c15a8cae64e2de9517fe8d471b5373/src/it/projects/MSHADE-467_parallel-dependency-reduced-pom/pom.xml#L64

With this PR, exclusions like the above (including `*`/`*`) are preserved in the reduced POM.

### What changed
- Core logic
  - Added `DependenciesExclusionResolver` to compute effective exclusions for both direct and transitive dependencies.
  - Introduced `DependencyList` utility to help build the reduced POM dependency set with proper scopes and exclusions.
  - Refactored `ShadeMojo` to apply resolved exclusions during reduced POM generation and to streamline transitive dependency handling.
- Tests
  - Added a new IT project: `dep-reduced-pom-with-transitive-dependencies-with-exclusions` (verifies propagation of direct and transitive exclusions).
  - Updated/added expected reduced POMs for MSHADE-467 parallel builds (`shadeMT1..shadeMT4/expected-dependency-reduced-pom.xml`).
  - Kept changes minimal for now to validate the approach first.

### Behavior before vs after
- Before: `dependency-reduced-pom.xml` could omit user-declared exclusions and omit required transitive exclusions, causing unexpected dependencies to appear for consumers.
- After: The reduced POM preserves original exclusions and includes necessary transitive exclusions, ensuring the effective dependency graph remains consistent.

### Why this matters
- Ensures consumers of the shaded artifact do not unintentionally pull dependencies that were explicitly excluded.
- Aligns the reduced POM with what users expect from the original POM’s dependency declarations.
- Improves reproducibility and clarity of dependency management for downstream builds.

### Testing
- Added basic integration tests to demonstrate the corrected behavior:
  - `src/it/projects/dep-reduced-pom-with-transitive-dependencies-with-exclusions/*`
  - Updated MSHADE-467 parallel ITs with `expected-dependency-reduced-pom.xml` snapshots for visibility into changes, for example:
    - `src/it/projects/MSHADE-467_parallel-dependency-reduced-pom/shadeMT1/expected-dependency-reduced-pom.xml`

**Note: At this stage only the foundational tests are included. If the overall approach is accepted, I will add more test coverage, including corner cases (wildcards, multiple excludes, ordering, and mixed scopes).**

### Open question: how strict should IT verification be?
Currently, to make differences obvious, I added snapshot files like:
- `src/it/projects/MSHADE-467_parallel-dependency-reduced-pom/shadeMT1/expected-dependency-reduced-pom.xml`

Is a strict one-to-one comparison (expected vs. generated reduced POM) desirable for ITs? My recommendation:
- Prefer structure-aware assertions over literal byte-for-byte equality to avoid false negatives due to formatting, ordering, or benign metadata changes. Options:
  - Normalize and compare: strip whitespace, sort `dependency` and `exclusion` lists where ordering is non-semantic, and compare normalized XML.
  - XPath-based checks: assert presence of specific dependencies and their exclusion blocks, and absence of those that should remain excluded.
- We can keep a small number of full-file snapshots for human readability, but rely on normalized or XPath checks for robustness.

I’m happy to implement whichever verification approach the project prefers.


src/it/projects/MSHADE-467_parallel-dependency-reduced-pom/shadeMT1/expected-dependency-reduced-pom.xml
```diff
@@ -0,0 +1,600 @@
+<!--
```
marwin1991: Preserve only for testing.

I’m wondering if I should add a 1:1 comparison between the generated reduced-pom.xml and the expected one (the expected file was generated in the main branch).

Then, in a pull request, maintainers could see the diff in the expected reduced-pom.xml, which would make it clearer how the pull request affects the plugin.

I’m a maintainer of a changelog plugin, and in that project I verify one-to-one changes for every integration test. This way, after community contributions, I can clearly see how they influence the final artifact—instead of just guessing what changed.

marwin1991:
Hello @slachiewicz @Bukama @slawekjaranowski @olamy can I somehow help you during review this changes?

PR #781, opened 2026-02-08
Update to parent POM v 47

Bukama:
Updates parent POM and dependencies, and fixes ITs to not use ancient plugins and Java versions.


Bukama:
> done via #785

Parts of it yes, other parts not.

PR #785, opened 2026-02-24
Bug: Extra JARs and Artifacts were not subjected to filtering

cstamas:
Bug: the `extraJars` and `extraArtifacts` were added to shaded JAR as is, without subjecting them to filtering.


olamy:
> If all ITs are busted then that should be fixed first in a separate PR.

not sure to understand the reasoning? what would be the added value of adding extra work/waste of time with a separate PR??

It has been discovered the reason of the failure with Java 25 so it's fixed now.

PR #787, opened 2026-02-26
Drop ASM; use already present one

cstamas:
Given `jdependency` is a **key dependency** of this plugin, and it already have full ASM shaded in, why not use it?

See
https://github.com/tcurdt/jdependency/issues/389

THIS IS JUST EXPERIMENT and pro-con discussion starter

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
+++ b/src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
@@ -66,13 +66,13 @@ import org.apache.maven.plugins.shade.resource.ReproducibleResourceTransformer;
 import org.apache.maven.plugins.shade.resource.ResourceTransformer;
 import org.codehaus.plexus.util.IOUtil;
 import org.codehaus.plexus.util.io.CachingOutputStream;
-import org.objectweb.asm.ClassReader;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/maven/plugins/shade/DefaultShader.java
```diff
@@ -66,13 +66,13 @@
...
 import org.codehaus.plexus.util.IOUtil;
 import org.codehaus.plexus.util.io.CachingOutputStream;
-import org.objectweb.asm.ClassReader;
-import org.objectweb.asm.ClassVisitor;
-import org.objectweb.asm.ClassWriter;
-import org.objectweb.asm.commons.ClassRemapper;
-import org.objectweb.asm.commons.Remapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.vafer.jdeb.shaded.objectweb.asm.ClassReader;
```
elharo: I don't know that we can rely on this being present. And has it been changed in any way?

elharo: But we can't rely on the shaded version of asm in jdependency being up to date with what we need. If we depend on asm directly, and we do, we should depend on it directly.

PR #827, opened 2026-07-29
[MSHADE-434] Restore project.getFile() after dependency-reduced POM creation

### LINKED ISSUE #453: [MSHADE-434] Strange side-effect on rat-plugin
**[Chesnay Schepler](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=zentol)** opened **[MSHADE-434](https://issues.apache.org/jira/browse/MSHADE-434?redirect=false)** and commented

When upgrading the shade-plugin from 3.2.4 to 3.3.0, without changing anything else, the Apache Flink builds are getting failed by the rat-plugin, complaining about a number of files in the root target/ directory.

These files do exist, but are usually excluded by the plugin.
For some reason upgrading the shade-plugin is changing this behavior.

```java
[WARNING] Files with unapproved licenses:
  /home/chesnay/dev/repos/flink/flink/target/rat.txt
  /home/chesnay/dev/repos/flink/flink/target/checkstyle-result.xml
  /home/chesnay/dev/repos/flink/flink/target/checkstyle-cachefile
```

This is reeeeally strange, and were wondering if you guys have any ideas for what could going on here.

Interestingly enough explicitly listing these files (e.g., `**/rat.txt`) results in them being excluded again.

This happened with Java 8 + Maven 3.2.5 and Java 11 + Maven 3.8.6 and is 100% reproducible.

---

**Affects:** 3.3.0

**Issue Links:**
- [MSHADE-321](https://issues.apache.org/jira/browse/MSHADE-321) Always respect 'createDependencyReducedPom' flag
 (_**"is caused by"**_)
- [FLINK-30083](https://issues.apache.org/jira/browse/FLINK-30083) Bump maven-shade-plugin to 3.4.1

elharo:
## Problem

Calling `project.setFile(dependencyReducedPomLocation)` in `ShadeMojo` changes the project file reference for all subsequent plugins in the build. This causes plugins like `apache-rat-plugin` and `maven-checkstyle-plugin` to resolve relative paths against the wrong directory, breaking their behavior. 

Specifically, when `dependencyReducedPomLocation` is set to a directory other than `${basedir}` (e.g., `target/`), `project.getFile()` returns the DRP path, and plugins relying on `getFile().getParentFile()` for basedir resolution will resolve paths incorrectly.

This was introduced/changed by MSHADE-321 (commit 6ea8543) which moved the `createDependencyReducedPom()` call outside of the inner artifact-processing block.

## Fix

Save `project.getFile()` before calling `createDependencyReducedPom()` and restore it afterward in `ShadeMojo.execute()`. This ensures `project.getFile()` continues to point to the original `pom.xml` after the shade plugin finishes, preventing side effects on subsequent plugins.

## Integration Test

`MSHADE-434_ratSideEffect`: Configures shade with `dependencyReducedPomLocation` in `target/`, then verifies via `maven-help-plugin:evaluate` that `${project.file}` still points to the original `pom.xml` after shading.

- Without fix: test fails (project.file is `dependency-reduced-pom.xml`)
- With fix: test passes (project.file is `pom.xml`)

All 72 unit tests pass.

Fixes #453

### DIFF
--- a/src/it/projects/MSHADE-434_ratSideEffect/pom.xml
+++ b/src/it/projects/MSHADE-434_ratSideEffect/pom.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.its.shade</groupId>
+  <artifactId>mshade-434</artifactId>
+  <version>1.0</version>
+  <packaging>jar</packaging>
+
+  <name>MSHADE-434</name>
+  <description>
+    Test that project.getFile() is restored to the original pom.xml after shading,
+    preventing side effects on subsequent plugins (e.g., RAT, checkstyle).
+  </description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven.its.shade.fac</groupId>
+      <artifactId>a</artifactId>
+      <version>0.1</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>@project.version@</version>
+        <executions>
+          <execution>
+            <id>shade</id>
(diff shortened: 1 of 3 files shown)


slachiewicz:
The red cell is this PR's own IT, not the change under test. `windows-latest jdk-8-zulu` is the only failing cell; the other 17 are green, and the invoker summary is `Passed: 83, Failed: 1` with `MSHADE-434_ratSideEffect\pom.xml` as the one failure.

`verify.groovy` looks for the `help:evaluate` output like this:

```groovy
def projectFileLine = lines.find { it =~ /^\// && it.endsWith("/pom.xml") }
assert projectFileLine != null
```

Both predicates assume a Unix path. On Windows the value is `D:\a\maven-shade-plugin\...\pom.xml`, which neither starts with `/` nor ends with `/pom.xml`, so `find` returns null and the assert fires before anything about `project.file` is actually checked:

    '/home/runner/work/p/pom.xml'  -> found = true
    'D:\a\...\pom.xml'             -> found = false

Matching on the file name instead of the separator fixes it — `it.trim().endsWith("pom.xml") && !it.trim().endsWith("dependency-reduced-pom.xml")` — or compare against `new File(basedir, "pom.xml").canonicalPath`, which is separator-correct by construction.

#828 carries the identical matcher and fails on the same single cell, so one fix covers both.

## On the overlap

#128, this PR, and #828 all address the same thing: `dependencyReducedPomLocation` leaving `project.getFile()` pointing at the reduced POM. #128 changes `setFile` to `setPomFile` so `basedir` is never reassigned; these two save the original and restore it afterwards. Both work, but they are alternatives rather than complements, and #128 has been approved and open since 2022. Worth a maintainer deciding which one lands before either is merged.

*This comment was created with AI assistance.*

PR #830, opened 2026-07-30
Use project name in generated Apache NOTICE

### LINKED ISSUE #742: `ApacheNoticeResourceTransformer` not injecting project name
### Affected version

3.6.0

### Bug description

When generating a `NOTICE` using `ApacheNoticeResourceTransformer`, the project name is not injected (or more likely, resolved) as expected, e.g. the output is:
> // ------------------------------------------------------------------
// NOTICE file corresponding to the section 4d of The Apache License,
// Version 2.0, in this case for
// ------------------------------------------------------------------

Not, something like this as you' might expect:
https://github.com/apache/amoro/blob/0e34e89dcfc8b5c5b5da285786004965d5b70bc6/NOTICE-binary#L1-L4

https://github.com/apache/maven-shade-plugin/blob/256eab83c008a1e701b16aad0b473c588a3560aa/src/main/java/org/apache/maven/plugins/shade/resource/ApacheNoticeResourceTransformer.java#L100

Reproducer:
Simple `pom.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>
<project
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="
         http://maven.apache.org/POM/4.0.0
         http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>com.me</groupId>
	<version>0.0.1</version>
	<artifactId>maven-shade-notice-reproducer</artifactId>
    <name>Maven Shade Notice Reproducer</name>
	<build>
		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-shade-plugin</artifactId>
				<version>3.6.0</version>
				<executions>
					<execution>
						<phase>package</phase>
						<goals>
							<goal>shade</goal>
						</goals>
						<configuration>
							<transformers>
								<transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer"/>
							</transformers>
						</configuration>
					</execution>
				</executions>
			</plugin>
		</plugins>
	</build>
	<dependencies>
		<dependency>
			<groupId>org.apache.commons</groupId>
			<artifactId>commons-lang3</artifactId>
			<version>3.18.0</version>
		</dependency>
	</dependencies>
</project>
```

Windows command to build and print the contents of the `NOTICE`:
`mvn clean package && jar -xf target\maven-shade-notice-reproducer-0.0.1.jar && type META-INF\NOTICE`

wilx:
## Summary

Use the Maven project name as the default project name in notices generated by
`ApacheNoticeResourceTransformer`.

The transformer previously initialized `projectName` to an empty string and the
mojo never supplied the injected `MavenProject`. Consequently, configuring the
transformer without an explicit `<projectName>` produced a blank generated
NOTICE header even when the POM had a nonblank `<name>`.

`ShadeMojo` now supplies `MavenProject#getName()` when the transformer project
name is unset and the Maven project name is neither null nor blank. Explicit
transformer configuration remains authoritative, and a missing or blank Maven
project name preserves the previous empty behavior.

The change includes unit coverage for the fallback rules, a failing-first
integration test for the generated NOTICE, and documentation of the default.

Fixes #742.

## Validation

- `mvn -Prun-its clean verify`: 74 unit tests and 84 integration projects
  passed; two projects were skipped by their JRE-version conditions.
- JDK 21 `mvn test`: 74 tests passed.
- Maven 4 built the isolated regression project with the branch plugin and
  generated the expected NOTICE header.
- `mvn site` passed.

       pulling in other changes.

       This change is tracked by GitHub issue #742; no separate JIRA issue was opened.
       commit message.

       The title uses `#742`, the repository's GitHub tracking issue.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
@@ -60,6 +60,7 @@ import org.apache.maven.plugins.shade.pom.PomWriter;
 import org.apache.maven.plugins.shade.relocation.Relocator;
 import org.apache.maven.plugins.shade.relocation.SerializedLambdaRelocator;
 import org.apache.maven.plugins.shade.relocation.SimpleRelocator;
+import org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer;
 import org.apache.maven.plugins.shade.resource.ManifestResourceTransformer;
 import org.apache.maven.plugins.shade.resource.ResourceTransformer;
 import org.apache.maven.project.DefaultProjectBuildingRequest;
@@ -970,6 +971,9 @@ public class ShadeMojo extends AbstractMojo {
             if (transformer == null) {
                 throw new MojoExecutionException(
                         "Failed to create shaded artifact: parameter transformers contains null (double-check XML attribute)");
+            }
+            if (transformer instanceof ApacheNoticeResourceTransformer) {
+                ((ApacheNoticeResourceTransformer) transformer).setProjectNameIfUnset(project.getName());
             }
         }
         return Arrays.asList(transformers);
--- a/src/main/java/org/apache/maven/plugins/shade/resource/ApacheNoticeResourceTransformer.java
+++ b/src/main/java/org/apache/maven/plugins/shade/resource/ApacheNoticeResourceTransformer.java
@@ -79,6 +79,19 @@ public class ApacheNoticeResourceTransformer extends AbstractCompatibilityTransf
 
     private static final String NOTICE_MD_PATH = "META-INF/NOTICE.md";
 
+    /**
+     * Uses the Maven project name when no project name was configured explicitly.
+     *
+     * @param projectName the Maven project name
+     */
+    public void setProjectNameIfUnset(String projectName) {
+        if (this.projectName.isEmpty()
+                && projectName != null
+                && !projectName.trim().isEmpty()) {
+            this.projectName = projectName;
+        }
+    }
+
     @Override
     public boolean canTransformResource(String resource) {
         return NOTICE_PATH.equalsIgnoreCase(resource)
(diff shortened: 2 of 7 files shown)


wilx:
Rebased.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -971,6 +972,9 @@ private List<ResourceTransformer> getResourceTransformers() throws MojoExecution
                 throw new MojoExecutionException(
                         "Failed to create shaded artifact: parameter transformers contains null (double-check XML attribute)");
             }
+            if (transformer instanceof ApacheNoticeResourceTransformer) {
```
elharo, line 975: might be necessary, but it would be cleaner if we could figure out some way to do this that doesn't require instanceof and casting.

src/main/java/org/apache/maven/plugins/shade/resource/ApacheNoticeResourceTransformer.java
```diff
@@ -79,6 +79,19 @@ public class ApacheNoticeResourceTransformer extends AbstractCompatibilityTransf
 
     private static final String NOTICE_MD_PATH = "META-INF/NOTICE.md";
 
+    /**
+     * Uses the Maven project name when no project name was configured explicitly.
+     *
+     * @param projectName the Maven project name
+     */
+    public void setProjectNameIfUnset(String projectName) {
```
elharo, line 87: why "if unset"? why not just set project name?

wilx: Because the field can be set via XML in the POM file, too. That takes precedence. If it is however not set via XML in POM, it will be set here. Thus if unset.

elharo: This still makes me queasy. My gut is that if someone calls this method, the project name should be set.

Even if we keep the behavior as it, we need a better name. setDefaultProjectName? setFallbackProjectName? maybe something else.

wilx: The current name is descriptive. It does not need a different name.

PR #843, opened 2026-08-14
Write an empty relativePath on dependency-reduced-pom.xml

### LINKED ISSUE #813: Set <relativePath/> in dependency-reduced-pom.xml to avoid false parent cycle in Maven 4
## Summary

The `dependency-reduced-pom.xml` generated by the shade plugin preserves the original POM's `<parent>` element without setting an explicit `<relativePath/>`. In Maven 4, this can cause a false parent cycle detection error when the default relative path `..` resolves to a POM with a different GA that shares the same parent reference.

## Reproduction

Build any project using the shade plugin with a parent POM, e.g., [apache/maven-doxia-converter](https://github.com/apache/maven-doxia-converter):

```
mvn package -DskipTests
```

With Maven 4, this fails with:
```
[FATAL] The parents form a cycle: org.apache.maven.doxia:doxia-tools:43 ->
  /path/to/project/pom.xml -> org.apache.maven.doxia:doxia-tools:43
  for project [inherited]:doxia-converter:jar:1.4-SNAPSHOT
  at /path/to/project/dependency-reduced-pom.xml
```

## Root cause

The generated `dependency-reduced-pom.xml` inherits the parent element as-is from the original POM. Since no `<relativePath/>` is set, Maven defaults to `..`, which may resolve to a POM that is not the actual parent. Maven 4's stricter model builder detects this as a cycle when the resolved POM shares the same parent reference.

## Proposed fix

When writing the `dependency-reduced-pom.xml`, set `<relativePath/>` (empty) in the parent element to explicitly disable local parent resolution. The parent should always be resolved from the repository for generated POMs.

## Workaround

A fix has been submitted for the Maven 4 model builder (apache/maven#12078, apache/maven#12079) that detects the GA mismatch early and avoids the false cycle. However, setting `<relativePath/>` in the generated POM is the correct long-term fix on the plugin side.

## Environment

- Maven: 4.0.0-rc-2+
- maven-shade-plugin: 3.6.0 / 3.6.1

Related: apache/maven#12074

_Claude Code on behalf of Guillaume Nodet_

AzazelSensei:

- [x] Your pull request should address just one issue, without pulling in other changes.
  Note that commits might be squashed by a maintainer on merge.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

Maven 4 treats a missing parent relativePath as `..`. The reduced POM
sits next to the project POM, so that default walks into a file that is
not the parent and reports a cycle.

I only keep a real relativePath when it actually points at the parent
POM. Otherwise I write an empty one so Maven goes to the repository.
Local-only parents still keep their path.

`PomWriterRelativePathTest.emptyRelativePathIsWritten` fails on current
master (the empty path is dropped). It passes here.

Fixes #813


src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -1260,6 +1269,23 @@ private void rewriteDependencyReducedPomIfWeHaveReduction(
         }
     }
 
+    /**
+     * Keep a local relativePath only when it really points at the parent POM.
+     * Otherwise Maven 4 treats the default {@code ..} as the parent and reports
+     * a cycle against a sibling POM that happens to share the same parent GAV.
+     */
+    static boolean parentFileIsTheConfiguredRelativePath(File configured, File actual) {
```
elharo: This method name has to be simplified.and should probably be private.

elharo (changes requested):
you might need to use an integration test to test this

AzazelSensei:
Renamed the helper to isSameFile. Left it package-private so the unit test can still call it.

src/test/java/org/apache/maven/plugins/shade/mojo/ShadeMojoParentPathTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class ShadeMojoParentPathTest {
+    @TempDir
+    File temporaryFolder;
+
+    @Test
+    public void matchingParentFilesKeepTheRelativePath() throws Exception {
+        File pom = new File(temporaryFolder, "pom.xml");
+        Files.write(pom.toPath(), "<project/>".getBytes("UTF-8"));
+        assertTrue(ShadeMojo.isSameFile(pom, pom));
```
elharo: This is a general issue in a lot of Maven projects. I should save this comment so I can copy and paste it in the future, or maybe make a blog post out of it because this comes up a lot. 

Making the method non-private allows it to be easily reached for unit tests. However it doesn't guarantee that the code in question is actually reached by Maven, or that it's called in the way you expect. 

An integration test should be able to test this without making anything non-private just for testing. If an integration test can't test this code, then it's very possible this code isn't needed at all.

LLMs are usually pretty good at contributing integration tests for Maven, as long as you prompt them to do that. Otherwise they tend to default to bad habits and shortcuts they've learned from existing code like making methods non-private just for testing.

AzazelSensei:
Dropped the isSameFile unit test and made the helper private. The PomWriter test still covers writing an empty relativePath.

src/test/java/org/apache/maven/plugins/shade/pom/PomWriterRelativePathTest.java
```diff
@@ -0,0 +1,72 @@
...
+        parent.setArtifactId("pa");
+        parent.setVersion("1.0");
+        parent.setRelativePath("");
+        model.setParent(parent);
+
+        StringWriter writer = new StringWriter();
+        PomWriter.write(writer, model, true);
+        String xml = writer.toString();
+
+        assertTrue(xml.contains("<relativePath"), xml);
```
elharo: could you expand this test to cover not just the tag but the complete element that shows the path?

elharo (changes requested):

AzazelSensei:
The empty path is now asserted as the full `<relativePath />` element. I also added a case for a custom path so the written value is checked, not just the tag.

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -1230,8 +1230,17 @@ private void rewriteDependencyReducedPomIfWeHaveReduction(
...
 
-                    String relPath = RelativizePath.convertToRelativePath(parentFile, f);
-                    model.getParent().setRelativePath(relPath);
+                    File actualParentFile =
+                            project.getParent() != null ? project.getParent().getFile() : null;
+                    if (isSameFile(parentFile, actualParentFile)) {
+                        String relPath = RelativizePath.convertToRelativePath(parentFile, f);
+                        model.getParent().setRelativePath(relPath);
+                    } else {
+                        // ../pom.xml is a different GAV (or missing). An empty
```
elharo: ../pom.xml has a different GAV (or does not have one).

src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java
```diff
@@ -1260,6 +1269,21 @@ private void rewriteDependencyReducedPomIfWeHaveReduction(
...
+     */
+    private static boolean isSameFile(File configured, File actual) {
+        try {
+            return actual != null
+                    && configured != null
+                    && actual.isFile()
+                    && configured.isFile()
+                    && actual.getCanonicalFile().equals(configured.getCanonicalFile());
+        } catch (IOException e) {
+            return false;
```
elharo: IIOException usually doesn't mean false. It's an exception that should be thrown

elharo (changes requested):

AzazelSensei:
isSameFile now throws IOException instead of treating it as a missing parent. Reworded the comment too.

elharo (approved):

PR #854, opened 2026-09-20
Improve log messages

michael-o:

- [ ] Your pull request should address just one issue, without pulling in other changes.
  Note that commits might be squashed by a maintainer on merge.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [ ] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).


slachiewicz (approved):

rmannibucau (approved):

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
```diff
@@ -191,12 +191,11 @@ private boolean removeServicesFromDir(Clazzpath cp, Set<Clazz> neededClasses, St
...
             }
 
-            try (BufferedReader configFileReader =
-                    new BufferedReader(new InputStreamReader(new FileInputStream(serviceProviderConfigFile), UTF_8))) {
+            try (BufferedReader configFileReader = Files.newBufferedReader(serviceProviderConfigFile.toPath(), UTF_8)) {
                 // check whether the found classes use services in turn
                 repeatScan |= scanServiceProviderConfigFile(cp, configFileReader);
             } catch (final IOException e) {
-                log.warn(e.getMessage());
+                log.warn("Failed to scan service provider config file " + serviceProviderConfigFile, e);
```
kriegaex, line 198: I would prefer
```java
log.warn( "Failed to scan service provider config file " + serviceProviderConfigFile );
log.debug( e );
```
Less noise in the `warn` log, full exception stack trace in the `debug` log.

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
```diff
@@ -222,11 +221,11 @@ private boolean removeServicesFromJar(Clazzpath cp, Set<Clazz> neededClasses, St
                     // check whether the found classes use services in turn
                     repeatScan = scanServiceProviderConfigFile(cp, configFileReader);
                 } catch (final IOException e) {
-                    log.warn(e.getMessage());
+                    log.warn("Failed to scan JAR entry " + jarEntry.getName(), e);
```
kriegaex, line 224: I would prefer
```java
log.warn( "Failed to scan JAR entry " + jarEntry + " in jar " + jar.getName() );
log.debug( e );
```
This also mentions the JAR name. Moreover, there is less noise in the `warn` log, but the full exception stack trace in the `debug` log.

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
```diff
@@ -222,11 +221,11 @@ private boolean removeServicesFromJar(Clazzpath cp, Set<Clazz> neededClasses, St
...
                 }
             }
         } catch (final IOException e) {
-            log.warn("Not a JAR file candidate. Ignoring classpath element '" + fileName + "' (" + e + ").");
+            log.warn("Not a JAR file candidate. Ignoring classpath element '" + fileName + "'", e);
```
kriegaex, line 228: Here, I also would rather split the warning and logging the exception stack trace, see above.

kriegaex (changes requested):
Of course, my review is non-binding, but you might want to consider the suggestions anyway. 🙂

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
rmannibucau: think it is saner to keep it together (or totally duplicate the line) cause loggers are not output streams, warn and debug can be routed to different output and there is no order delivery guarantee between them if executed super fast

kriegaex: @rmannibucau I thought that the log channels also correspond to log levels, i.e. that e.g. `debug` is a superset of `warn` with more detail. Is that assumption false? Is it really necessary to force users to see stack traces in the `warn` log?

michael-o: both messages are disjoint and that is not OK to print it like that, moreove it would need to consistent with the rest of the log statements which it is not.

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
michael-o: No, see above.

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
michael-o: Agreed on the message, but not on the second statement.

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
rmannibucau: @kriegaex it is not a superset just a plain hierarchy but you cant assume 2 disjoint messages will be linked if there is no explicit link between there like there

kriegaex (approved):
If the regular committers in the project disagree with my suggestions, there is nothing I can do. I can live with the result of this PR. Thank you so much for taking the time to open and process it upon my request. I do appreciate that. 🙂

src/main/java/org/apache/maven/plugins/shade/filter/MinijarFilter.java
michael-o: What can be done is:
```
if (debug enabled)
log.warn(message, e)
else
log.warn(message)
```

This pattern we us in several places.

rmannibucau:
```
if (debug enabled)
log.warn(message, e)
else
log.warn(message + "'" + e.getMessage() + "'")
```

maybe (or anything formatting you like) to not loose the error message which is often the key to understand the error - less critical when you have the real stack/causes

